PR #6, opened 2022-01-08
[ci] setup mingw CI

inkydragon:
+ setup mingw CI: Mingw-x64-static-release = 1 build
+ ubuntu = [static, shared] = 2 build
+ macos = [static, shared] = 2 build


cody271 (approved):
LGTM

PR #23, opened 2022-01-14
[Old PR libui#512] Add new API functions to get and set the visibility of table headers.

szanni:
Old PR  [libui#512](https://github.com/andlabs/libui/pull/512) 

Proposal for two new API additions to un/hide table headers:

```c
int uiTableHeaderVisible(uiTable *t);
void uiTableHeaderSetVisible(uiTable *t, int visible);
```

This was one of the requests in the table master issue.

--

Added API functions:
uiTableHeaderVisible() to determine whether the table header is visible.
uiTableHeaderSetVisible() to set the visibility of the table header.

Implementation provided for unix, darwin, and windows.

Notes: as darwin does not provide an API for hiding or recreating the
table header I opted for saving a reference and restoring that when the
visibility is set back to true. Setting the header to nil to hide it is
the suggested method for hiding the header according to the [docs](https://developer.apple.com/documentation/appkit/nstableview/1535880-headerview).

### DIFF
--- a/windows/table.cpp
+++ b/windows/table.cpp
@@ -479,6 +479,26 @@ void uiTableAppendButtonColumn(uiTable *t, const char *name, int buttonModelColu
 	p->buttonClickableModelColumn = buttonClickableModelColumn;
 }
 
+int uiTableHeaderVisible(uiTable *t)
+{
+	HWND header = ListView_GetHeader(t->hwnd);
+	if (header) {
+		LONG style = GetWindowLong(header, GWL_STYLE);
+		return !(style & HDS_HIDDEN);
+	}
+	// TODO abort here?
+	return 0;
+}
+
+void uiTableHeaderSetVisible(uiTable *t, int visible)
+{
+	LONG style = GetWindowLong(t->hwnd, GWL_STYLE);
+	if (visible)
+		SetWindowLong(t->hwnd, GWL_STYLE, style & ~LVS_NOCOLUMNHEADER);
+	else
+		SetWindowLong(t->hwnd, GWL_STYLE, style | LVS_NOCOLUMNHEADER);
+}
+
 uiTable *uiNewTable(uiTableParams *p)
 {
 	uiTable *t;
--- a/windows/table.cpp
+++ b/windows/table.cpp
@@ -481,12 +481,12 @@ void uiTableAppendButtonColumn(uiTable *t, const char *name, int buttonModelColu
 
 int uiTableHeaderVisible(uiTable *t)
 {
-	HWND header = ListView_GetHeader(t->hwnd);
+	HWND header =  (HWND) SendMessageW(t->hwnd, LVM_GETHEADER, 0, 0);
 	if (header) {
 		LONG style = GetWindowLong(header, GWL_STYLE);
 		return !(style & HDS_HIDDEN);
 	}
-	// TODO abort here?
+	uiprivImplBug("window handle %p unknown error from send LVM_GETHEADER", t->hwnd);
 	return 0;
 }
 
--- a/darwin/table.m
+++ b/darwin/table.m
@@ -14,8 +14,10 @@
 @interface uiprivTableView : NSTableView {
 	uiTable *uiprivT;
 	uiTableModel *uiprivM;
+	NSTableHeaderView *headerViewRef;
 }
 - (id)initWithFrame:(NSRect)r uiprivT:(uiTable *)t uiprivM:(uiTableModel *)m;
+- (void)restoreHeaderView;
 @end
 
 @implementation uiprivTableView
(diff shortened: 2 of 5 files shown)


windows/table.cpp
```diff
@@ -479,6 +479,26 @@ void uiTableAppendButtonColumn(uiTable *t, const char *name, int buttonModelColu
 	p->buttonClickableModelColumn = buttonClickableModelColumn;
 }
 
+int uiTableHeaderVisible(uiTable *t)
+{
+	HWND header = ListView_GetHeader(t->hwnd);
```
cody271, line 484:
```suggestion
	HWND header = (HWND) SendMessageW(t->hwnd, LVM_GETHEADER, 0, 0);
```

windows/table.cpp
```diff
@@ -479,6 +479,26 @@ void uiTableAppendButtonColumn(uiTable *t, const char *name, int buttonModelColu
...
+	if (header) {
+		LONG style = GetWindowLong(header, GWL_STYLE);
+		return !(style & HDS_HIDDEN);
+	}
+	// TODO abort here?
```
cody271, line 489:
```suggestion
	uiprivImplBug("window handle %p unknown error from send LVM_GETHEADER", t->hwnd);
```

cody271 (commented):
`windows`: [`LVM_GETHEADER`](https://docs.microsoft.com/en-us/windows/win32/controls/lvm-getheader) message appears to not set an error, so not much to do besides report "unknown error" back to the user from `uiTableHeaderVisible()`. Also try to match the existing style of explicitly sending a WM message, instead of using the wrapper. The wrappers are less portable across different Windows toolchains.

cody271:
Also wrapper macros [here](https://github.com/libui-ng/libui-ng/commit/595e5a5b5f71518960e15481337dae764aeb85f5#diff-340d600392818df2413382dc7d8325c360d83ea49a262d31760348484bbc10b5)

windows/table.cpp
szanni: Great, thank you. The `uiprivImplBug` function was exactly what I was looking for, did not know it existed.

szanni:
> `windows`: [`LVM_GETHEADER`](https://docs.microsoft.com/en-us/windows/win32/controls/lvm-getheader) message appears to not set an error, so not much to do besides report "unknown error" back to the user from `uiTableHeaderVisible()`. Also try to match the existing style of explicitly sending a WM message, instead of using the wrapper. The wrappers are less portable across different Windows toolchains.

I did not know the macros were less portable. Any reference for that? Because the WM messages look hideous to me. So much clutter and risk of introducing bugs that I would personally love to avoid.
I was actually going to suggest enforcing the use of the wrapper macros for newer code.

cody271:
> I did not know the macros were less portable. Any reference for that? Because the WM messages look hideous to me. So much clutter and risk of introducing bugs that I would personally love to avoid. I was actually going to suggest enforcing the use of the wrapper macros for newer code.
>
Hrmm the references I'm finding to this issue are [very old](https://sourceforge.net/p/mingw/mailman/message/12581475/). I would hope it _is_ fixed by now, decade(s) later.
Some references to this issue in our `windows/` code already [here](https://github.com/libui-ng/libui-ng/blob/master/windows/uipriv_windows.hpp#L57).

cody271:
> Because the WM messages look hideous to me. So much clutter and risk of introducing bugs that I would personally love to avoid. I was actually going to suggest enforcing the use of the wrapper macros for newer code.
>
You are not wrong, it is hideous. Regardless though, the more important point is that we should stay as _consistent_ as possible with all the existing conventions. When we do adopt the macro wrappers over sending WM messages, it needs to be all or nothing across the entire `windows/` codebase.

szanni:
> Hrmm the references I'm finding to this issue are [very old](https://sourceforge.net/p/mingw/mailman/message/12581475/). I would hope it _is_ fixed by now, decade(s) later. Some references to this issue in our `windows/` code already [here](https://github.com/libui-ng/libui-ng/blob/master/windows/uipriv_windows.hpp#L57).

That does indeed look old and seems to refers to mingw, which to my knowledge has always been _very_ incomplete. I am pretty sure that's why only mingw-w64 is listed as supported in the first place.

> You are not wrong, it is hideous. Regardless though, the more important point is that we should stay as _consistent_ as possible with all the existing conventions. When we do adopt the macro wrappers over sending WM messages, it needs to be all or nothing across the entire `windows/` codebase.

Ok. I can change things to use SendMessageW. My idea was to use macros in new code while leaving SendMessageW in old code so we don't get too many merge conflicts.
I guess changing things could be done at a later stage though, just to make sure we don't break anything in the fork for the time being at least.
It is just code like [this](https://github.com/libui-ng/libui-ng/blob/master/windows/table.cpp#L485) that seems very unsafe to me and rife for bugs.

cody271:
> 
> I guess changing things could be done at a later stage though, just to make sure we don't break anything in the fork for the time being at least.
> It is just code like [this](https://github.com/libui-ng/libui-ng/blob/master/windows/table.cpp#L485) that seems very unsafe to me and rife for bugs.
>
Issue to fix this [here](https://github.com/libui-ng/libui-ng/issues/35)

cody271 (approved):
LGTM

PR #24, opened 2022-01-15
[Old PR libui#513] Add new API functions for table header clicks and sort indicators.

szanni:
Old PR  [libui#513](https://github.com/andlabs/libui/pull/513) 

Proposal for introducing an enum sort type `uiSort`:
```c
uiSortIndicatorNone
uiSortIndicatorAscending
uiSortIndicatorDescending
```
Two functions for setting table header sort indicators (only visual, no sorting is performed):
```c
uiSort uiTableHeaderSortIndicator(uiTable *t, int column);
void uiTableHeaderSetSortIndicator(uiTable *t, int column, uiSortIndiactor indicator);
```
And a setter for a header on clicked callback function:
```c
void uiTableHeaderOnClicked(uiTable *t, void (*f)(uiTable *table, int column, void *data), void *data);
```

Implementations are provided for darwin, unix, and windows.

Notes: The column is the index of the column when it was added to the table. It might be nice to be able to set a numeric identifier, similar to how table columns are handled on darwin. Or possibly even introducing a `uiTableColumn` type and cleaning up the `uiTableAppendColumn` API. This would however break the existing API, so I left it out for the time being.

### DIFF
--- a/darwin/table.h
+++ b/darwin/table.h
@@ -16,6 +16,8 @@ struct uiTable {
 	uiprivScrollViewData *d;
 	int backgroundColumn;
 	uiTableModel *m;
+	void (*headerOnClicked)(uiTable *, int, void *);
+	void *headerOnClickedData;
 };
 
 // tablecolumn.m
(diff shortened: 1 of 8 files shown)


szanni:
Just updated the branch renaming the type and members to `uiSortIndicator*` and reverted back to using the `SendMessageW` function instead of macros.

cody271 (approved):
LGTM

PR #34, opened 2022-01-16
New examples

inkydragon:
- `hello-world`: Display a window. Using as few statements as possible may not be the best practice.
- `hello-window`: Display a window and add a label to the window, using best practices where possible.

TODO: Add more standalone `hello-widgets` examples.


cody271:
Aren't these two new examples the same?

cody271:
> Aren't these two new examples the same?
>
(Nevermind. I see one creates a label in the window.)

Maybe the examples should be something like `hello-window-1.c` and `hello-label-2.c`, to make it clear they are part of a series of examples?

cody271:
Also we could add `doc/examples.md` to link all the examples together from the Doxygen main page, including some screenshots of each example with the code.

inkydragon:
> Maybe the examples should be something like hello-window-1.c and hello-label-2.c, to make it clear they are part of a series of examples?

Yes! For some simple examples, a single c file may be enough, no need to create a separate folder.

My proposal:
- `screenshots/`: All in one directory for centralized referencing.
- `widgets/`
  - hello-label.c
  - `hello-window/`
    - hello-window-1.c
    - hello-window-2.c
  - `hello-xxx/`
  - hello-xxx.c

----

> including some screenshots of each example with the code.

We would write a simple example for each widget, so that there would be a lot of images.

Should we put all the images in the library here?

If it's hard to decide in a short time. 
We can set an upper size limit for all images, for example: 10 MB, and wait until all images are over the limit, then split them into separate libraries.

cody271:
> 
> Yes! For some simple examples, a single c file may be enough, no need to create a separate folder.
> 
Yup single C file per example is totally fine. We should put the whole example series under maybe `examples/tutorial`, something that's easy to find. Then GitHub will automatically load `examples/tutorial/README.md`.

>
> Should we put all the images in the library here?
> 
> If it's hard to decide in a short time. We can set an upper size limit for all images, for example: 10 MB, and wait until all images are over the limit, then split them into separate libraries.
>
I think its okay for now, we can move it later if they are too big. 10MB is probably a good limit. We already have 188KB of images for one example.
```sh
libui-ng$ du -hc examples/controlgallery/*.png
 96K	examples/controlgallery/darwin.png
 44K	examples/controlgallery/unix.png
 48K	examples/controlgallery/windows.png
188K	total
```

szanni:
> Maybe the examples should be something like `hello-window-1.c` and `hello-label-2.c`, to make it clear they are part of a series of examples?

I would definitely put label in the file name. Makes it much clearer and more discoverable.

Apart from that: if we should add more screenshots I believe we should optimize them before submitting them upstream.
I was able to reduce the size of the existing screenshots by ~25% by running them through `optipng -o7` and `zopflipng -m`. Just so we bloat the repo as little as possible.

```
% du -hc examples/controlgallery/*.png                                                                                                                                                                                                
65K	examples/controlgallery/darwin.png
37K	examples/controlgallery/unix.png
41K	examples/controlgallery/windows.png
142K	total
```

cody271:
> 
> I believe we should optimize them before submitting them upstream.
> Just so we bloat the repo as little as possible.
> 
Absolutely. 💯 

Repo needs to remain small enough to clone quickly!

examples/hello-window/main.c
```diff
@@ -0,0 +1,39 @@
+#include <stdlib.h> // exit
+#include <string.h> // memset
+#include <stdio.h> // fprintf
+#include "../../ui.h"
+
+
+int onClosing(uiWindow *w, void *data)
+{
+    uiQuit();
```
szanni, line 9: Indentation throughout the project is 1 tab. I believe this should be enforced in the examples as well.

examples/hello-world/main.c
```diff
@@ -0,0 +1,27 @@
...
+    memset(&o, 0, sizeof (uiInitOptions));
+    if (uiInit(&o) != NULL) exit(1);
+
+    // Create a new window
+    w = uiNewWindow("Hello, World!", 300, 30, 0);
+
+    uiWindowOnClosing(w, onClosing, NULL);
+    uiControlShow(uiControl(w));
+    uiMain();
+    return 0;
```
szanni, line 26: Missing 
```suggestion
    uiUninit();
    return 0;
```

examples/hello-world/main.c
```diff
@@ -0,0 +1,27 @@
...
+    return 1;
+}
+
+int main(void)
+{
+    uiInitOptions o;
+    uiWindow *w;
+
+    memset(&o, 0, sizeof (uiInitOptions));
+    if (uiInit(&o) != NULL) exit(1);
```
szanni, line 18: Do we want to make error handling consistent across examples? Why not error printing here?

szanni (changes requested):
Apart from the small suggestions and missing `uiUninit`, I would rename the files.
`hello-world` should print "Hello World!" in the window. As a user I would expect the two files to contain the exact opposite.
`hello-window`: creates a window
`hello-world`: creates a window printing hello world

Otherwise the other naming suggestions would be fine too I guess. But I probably would remove the _hello_ in that case. Just name them `window`, `label`, etc. And the current `hello-window` as a `hello-world`, just because of the custom.

examples/hello-world/main.c
inkydragon: I just want to keep this example as simple as possible.

> make error handling consistent across examples

This is a good point. So I'll add error handling.

examples/hello-world/main.c
```diff
@@ -0,0 +1,35 @@
+#include <stdlib.h> // exit
```
szanni, line 1: Can be removed as there is not call to exit() anymore.

examples/window/main.c
```diff
@@ -0,0 +1,39 @@
+#include <stdlib.h> // exit
```
szanni, line 1: Can be removed as there is not call to exit() anymore.

szanni (changes requested):
Very nice. I believe we can remove the two includes. And I would still swap the two file names. I am greeted with an empty window on my system when running hello world. I do not see the text `hello world` which is the custom.

mohad12211:
Now that we have good documentation, unit testing, a11y and [QA](https://github.com/libui-ng/libui-ng/pull/133) soon, what's missing are new simple examples for most controls with screenshots, because the current `controlgallery` and `histogram` don't appear welcoming for newcomers at all. 

> Maybe the examples should be something like hello-window-1.c and hello-label-2.c, to make it clear they are part of a series of examples?

This looks good.

>  And I would still swap the two file names

yes, I think `hello-window` should show a window, `hello-world` should show hello world, inside a label.

examples/hello-world/main.c
```diff
@@ -0,0 +1,35 @@
...
+#include <string.h> // memset
```
szanni, line 2:
```suggestion
```

examples/hello-world/main.c
szanni, line 1:
```suggestion
```

examples/hello-world/main.c
```diff
@@ -0,0 +1,35 @@
...
+#include <stdio.h> // fprintf
+#include "../../ui.h"
```
szanni, line 4:
```suggestion
#include <ui.h>
```

So this can virtually be dropped into place.
We might need to tell meson to add the top level directory to the include path? Not sure, but this may already be the case.

examples/hello-world/main.c
```diff
@@ -0,0 +1,35 @@
...
+
+int onClosing(uiWindow *w, void *data)
+{
+	uiQuit();
+	return 1;
+}
+
+int main(void)
+{
+	uiInitOptions o;
```
szanni, line 15:
```suggestion
	uiInitOptions o = {0};
```

examples/hello-world/main.c
```diff
@@ -0,0 +1,35 @@
...
+	const char *err;
+	uiWindow *w;
+
+	memset(&o, 0, sizeof (uiInitOptions));
```
szanni, line 19:
```suggestion
```

examples/window/main.c
```diff
@@ -0,0 +1,39 @@
...
+#include <string.h> // memset
```
szanni, line 2:
```suggestion
```

examples/window/main.c
```diff
@@ -0,0 +1,39 @@
...
+#include <stdio.h> // fprintf
+#include "../../ui.h"
```
szanni, line 4:
```suggestion
#include <ui.h>
```

examples/window/main.c
```diff
@@ -0,0 +1,39 @@
...
+
+int onClosing(uiWindow *w, void *data)
+{
+	uiQuit();
+	return 1;
+}
+
+int main(void)
+{
+	uiInitOptions o;
```
szanni, line 15:
```suggestion
	uiInitOptions o = {0};
```

examples/window/main.c
```diff
@@ -0,0 +1,39 @@
...
+	const char *err;
+	uiWindow *w;
+	uiLabel *label;
+
+	memset(&o, 0, sizeof (uiInitOptions));
```
szanni, line 20:
```suggestion
```

szanni (changes requested):
Switch the file names around. Not all systems display window titles. The `hello-world` program just displays an empty window for me.

PR #37, opened 2022-01-18
 [Old PR libui#424] Make slider tool tip default on all platforms. Add API functions to enable/disable slider tool tips.

szanni:
Old PR [libui#424](https://github.com/andlabs/libui/pull/424)

As can be seen in the README screenshots the slider is not consistent across platforms.

This PR does:
- Disable value drawn over the slider on unix platforms
- Draw tool tip on unix
- Draw tool tip on windos
- Keep existing tooltip on windows
- Add API function to enable and disable tool tips.

```c
int uiSliderHasToolTip(uiSlider *s);
void uiSliderSetHasToolTip(uiSlider *s, int hasToolTip);
```

Tested on Linux, macOS and wine/mingw-w64.

### DIFF
--- a/darwin/slider.m
+++ b/darwin/slider.m
@@ -26,7 +26,10 @@ struct uiSlider {
 	NSSlider *slider;
 	void (*onChanged)(uiSlider *, void *);
 	void *onChangedData;
+	bool hasToolTip;
 };
+
+static void _uiSliderUpdateToolTip(uiSlider *s);
 
 @interface sliderDelegateClass : NSObject {
 	uiprivMap *sliders;
@@ -58,6 +61,9 @@ struct uiSlider {
 
 	s = (uiSlider *) uiprivMapGet(self->sliders, sender);
 	(*(s->onChanged))(s, s->onChangedData);
+
+	if (s->hasToolTip)
+		_uiSliderUpdateToolTip(s);
 }
 
 - (void)registerSlider:(uiSlider *)s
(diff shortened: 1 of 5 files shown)


darwin/slider.m
```diff
@@ -26,8 +26,11 @@ - (NSSize)intrinsicContentSize
 	NSSlider *slider;
 	void (*onChanged)(uiSlider *, void *);
 	void *onChangedData;
+	bool hasToolTip;
```
cody271, line 29:
```suggestion
	BOOL hasToolTip;
```
Nit: use `BOOL` here because its Objective-C, I'll fix this.

unix/slider.c
```diff
@@ -1,6 +1,12 @@
 // 11 june 2015
 #include "uipriv_unix.h"
 
+/* Number of digits to represent an n-bit number in decimal: log10(2^n)
+ * Simplify, truncate & add 1 for full digits, add 1 for negative numbers:
+ *   -> n log10(2) + 2   <=   n * 0.302 + 2
+ */
+#define MAX_STRLEN_FOR_NBITS_IN_DECIMAL(n) ((int)((n) * 302 / 1000) + 2)
```
cody271, line 8: Just for reference in case anybody was wondering: this is only 12 bytes for the buffer, including NULL terminator.

cody271 (approved):
One quirk I noticed in testing: tooltip shows upon mouse hover on `unix`/`darwin`, but on `windows` it requires a mouse click first. This appears to be the expected (and only available) behavior though, nothing to do.

PR #39, opened 2022-01-19
Add AUTHORS.md

cody271:


szanni:
Looks good in general. Do we want to include e-mail? Projects like vlc [don't include e-mail](https://github.com/videolan/vlc/blob/master/AUTHORS).

And I would try to use the real names over the nick wherever possible.

cody271:
> Looks good in general. Do we want to include e-mail? Projects like vlc [don't include e-mail](https://github.com/videolan/vlc/blob/master/AUTHORS).
> 
Updated with just names, and fixed duplicate.

cody271:
> Projects like vlc
> 
[Nice](https://github.com/videolan/vlc/blob/master/AUTHORS#L184) 😎

cody271:
@szanni Just reworked this PR with a couple improvements:

a) Authors file is now automatically generated.
No reason to be manually editing this at all. New script uses data from GitHub API.
See `docs/build_authors_md.py`.

b) Use GitHub as the main contact point for contributors.
Thinking about what's actually important to keep track of, from a compliance perspective, is reliable contact information. Not their name or an email used in commits from years back. GitHub is more useful for this long term. (For example: if there ever is a copyright dispute, and you can't actually find a contributor.. you risk having to immediately delete all of their past contributions.)

cody271:
@szanni Merging this if there are no objections..

szanni:
The script is great, I guess it still does require manual oversight though?
Users deleting the GH is something that comes to mind.

cody271:
> Users deleting the GH is something that comes to mind.

From a compliance perspective, GitHub doesn't [handle this](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-user-account-settings/deleting-your-user-account) very well honestly. (Everything gets assigned to the user @ghost.)
An example in our old PRs list: https://github.com/andlabs/libui/pull/373. Who actually wrote that? We don't know.

The authors file script actually helps us here though: since we're caching the GitHub data locally inside the file, we can preserve authorship information from the deleted accounts. For now, I will just manually not merge any such changes. In the future, it could do this automatically: enforce that authors file changes are append only.

PR #43, opened 2022-01-19
[Old PR libui#514] Add new API functions to get/set table column widths. #514 

szanni:
Old PR [libui#514](https://github.com/andlabs/libui/pull/514)

Proposal to add API functions to get and set table column widths:
```c
int uiTableColumnWidth(uiTable *t, int column);
void uiTableColumnSetWidth(uiTable *t, int column, int width);
```
Same as in #24 we need a way to identify columns. I used the same mechanism of identifying columns by index position when added for the time being. 

Implementations provided for all platforms (darwin, unix, windows) with a test bed in page16.

### DIFF
--- a/windows/table.cpp
+++ b/windows/table.cpp
@@ -588,10 +588,10 @@ uiTable *uiNewTable(uiTableParams *p)
 
 int uiTableColumnWidth(uiTable *t, int column)
 {
-	return ListView_GetColumnWidth(t->hwnd, column);
+	return SendMessageW(t->hwnd, LVM_GETCOLUMNWIDTH, (WPARAM) column, 0);
 }
 
 void uiTableColumnSetWidth(uiTable *t, int column, int width)
 {
-	ListView_SetColumnWidth(t->hwnd, column, width);
+	SendMessageW(t->hwnd, LVM_SETCOLUMNWIDTH, (WPARAM) column, (LPARAM) width);
 }
--- a/windows/table.cpp
+++ b/windows/table.cpp
@@ -585,3 +585,13 @@ uiTable *uiNewTable(uiTableParams *p)
 
 	return t;
 }
+
+int uiTableColumnWidth(uiTable *t, int column)
+{
+	return ListView_GetColumnWidth(t->hwnd, column);
+}
+
+void uiTableColumnSetWidth(uiTable *t, int column, int width)
+{
+	ListView_SetColumnWidth(t->hwnd, column, width);
+}
--- a/windows/table.cpp
+++ b/windows/table.cpp
@@ -593,5 +593,8 @@ int uiTableColumnWidth(uiTable *t, int column)
 
 void uiTableColumnSetWidth(uiTable *t, int column, int width)
 {
+	if (width == -1)
+		width = LVSCW_AUTOSIZE_USEHEADER;
+
 	SendMessageW(t->hwnd, LVM_SETCOLUMNWIDTH, (WPARAM) column, (LPARAM) width);
 }
--- a/darwin/table.m
+++ b/darwin/table.m
@@ -286,3 +286,15 @@ void uiTableHeaderSetSortIndicator(uiTable *t, int lcol, uiSortIndicator indicat
 		img = nil;
 	[t->tv setIndicatorImage:img inTableColumn:tc];
 }
+
+int uiTableColumnWidth(uiTable *t, int column)
+{
+	NSTableColumn *tc = [t->tv tableColumnWithIdentifier:[@(column) stringValue]];
+	return [tc width];
+}
+
+void uiTableColumnSetWidth(uiTable *t, int column, int width)
+{
+	NSTableColumn *tc = [t->tv tableColumnWithIdentifier:[@(column) stringValue]];
+	[tc setWidth: width];
+}
--- a/darwin/table.m
+++ b/darwin/table.m
@@ -296,5 +296,10 @@ int uiTableColumnWidth(uiTable *t, int column)
 void uiTableColumnSetWidth(uiTable *t, int column, int width)
 {
 	NSTableColumn *tc = [t->tv tableColumnWithIdentifier:[@(column) stringValue]];
-	[tc setWidth: width];
+
+	if (width == -1)
+		//TODO: resize not only to header but also to max content width
+		[tc sizeToFit];
+	else
+		[tc setWidth: width];
 }
--- a/test/page16.c
+++ b/test/page16.c
@@ -105,6 +105,20 @@ void headerVisibleToggled(uiCheckbox *c, void *data)
 	uiCheckboxSetChecked(c, uiTableHeaderVisible(t));
 }
 
+uiSpinbox *columnID;
+uiSpinbox *columnWidth;
+static void changedColumnID(uiSpinbox *s, void *data)
+{
+	uiTable *t = data;
+	uiSpinboxSetValue(columnWidth, uiTableColumnWidth(t, uiSpinboxValue(columnID)));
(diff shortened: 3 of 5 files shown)


windows/table.cpp
```diff
@@ -585,3 +585,13 @@ uiTable *uiNewTable(uiTableParams *p)
...
 }
+
+int uiTableColumnWidth(uiTable *t, int column)
+{
+	return ListView_GetColumnWidth(t->hwnd, column);
+}
+
+void uiTableColumnSetWidth(uiTable *t, int column, int width)
+{
+	ListView_SetColumnWidth(t->hwnd, column, width);
```
cody271, line 596:
```suggestion
        // -1 to unset column width
        if (width == -1)
		width = LVSCW_AUTOSIZE_USEHEADER;
	ListView_SetColumnWidth(t->hwnd, column, width);
```

cody271 (commented):
This should handle _unsetting_ the fixed column width as well.

windows/table.cpp
szanni: I'll check how/if unsetting is possible on the other platforms and document it in the API.

And sorry I missed the macro usage again. I'll fix that as soon as I find some time.

szanni:
## From an API perspective:

Do we want to use a special value `-1` to unset the width? Or would it be preferable to have a separate API call for that?
Like `uiTableColumnSizeToFit()` `uiTableColumnResetWidth()`   `uiTableColumnUnsetWidth()` `uiTableColumnAutoSize()` ?

## Form the docs:

**windows**
LVSCW_AUTOSIZE: Automatically sizes the column.
LVSCW_AUTOSIZE_USEHEADER: Automatically sizes the column to fit the header text. If you use this value with the last column, its width is set to fill the remaining width of the list-view control.

**darwin**
sizeToFit(): Resizes the table column to fit the width of its header cell.

**unix**
If fixed_width is not -1, sets the fixed width of tree_column ; otherwise unsets it.

## Testing the three platforms I discovered that:

**windows** resizes to header width or content - which ever one is bigger (using `LVSCW_AUTOSIZE_USEHEADER`, this is contrary to what the docs say)
**darwin** resizes to header width (using `sizeToFit`)
**unix** resizes to header width or content - which ever one is bigger 

So we have the same behavior on windows and unix but not on darwin.

I could not find a function on darwin to set the width to either the header width or content width (which ever one is bigger) to match the behavior on windows and unix. There [seems to be a way](https://stackoverflow.com/questions/4674163/nstablecolumn-size-to-fit-contents) of doing so manually by iterating through all rows manually, finding the biggest and setting the width to that. Or save the original width and reset to that?

Different platforms handle the editing of cells differently as well. Unix will auto resize to fit content. Darwin will not.

I am not sure what the best course of action would be. Looking for feedback.

cody271:
> Do we want to use a special value `-1` to unset the width? Or would it be preferable to have a separate API call for that?
> Like `uiTableColumnSizeToFit()` `uiTableColumnResetWidth()` `uiTableColumnUnsetWidth()` `uiTableColumnAutoSize()` ?

I think just using `-1` to mean unset here is most in line with our minimalist vision..

For clarity, we can _also_ add a new single value enum:
```c
_UI_ENUM(uiTableColumnWidthHint) {
    uiTableColumnWidthHintAutoSize = -1,
};
```

cody271:
> to match the behavior on windows and unix. There [seems to be a way](https://stackoverflow.com/questions/4674163/nstablecolumn-size-to-fit-contents) of doing so manually by iterating through all rows manually, finding the biggest and setting the width to that.

Yup this is the only way to do it on darwin I think.

szanni:
> I think just using `-1` to mean unset here is most in line with our minimalist vision..
> 
> For clarity, we can _also_ add a new single value enum:

I would personally not necessarily add an enum. That will just upset the type checker.

> > to match the behavior on windows and unix. There [seems to be a way](https://stackoverflow.com/questions/4674163/nstablecolumn-size-to-fit-contents) of doing so manually by iterating through all rows manually, finding the biggest and setting the width to that.
> 
> Yup this is the only way to do it on darwin I think.

[Apparently](https://stackoverflow.com/questions/4674163/nstablecolumn-size-to-fit-contents#comment22815194_4767641) darwin will randomly sample rows for their widths on bigger tables. I am personally interested in using the library with tables with 10-20k+ rows. I guess we should try to replicate that behavior?

cody271:
> I would personally not necessarily add an enum. That will just upset the type checker.

Yes 😛 I guess I really mean "pre-defined int value with nice name".

```c
#define uiTableColumnWidthHintAutoSize        -1
```

cody271:
> [Apparently](https://stackoverflow.com/questions/4674163/nstablecolumn-size-to-fit-contents#comment22815194_4767641) darwin will randomly sample rows for their widths on bigger tables. I am personally interested in using the library with tables with 10-20k+ rows. I guess we should try to replicate that behavior?

Most likely, yes. First step would be adding a new table "stress test" page that creates a table that large.

szanni:
> Most likely, yes. First step would be adding a new table "stress test" page that creates a table that large.

I've been doing that locally but purposefully not pushing that upstream though due to memory usage concerns.
Just tested 20k rows in page17, it seems to make very little difference memory usage wise.

szanni:
I pushed a few more commits adding `-1` as special value. Setting `-1` will:
windows: set column width to header <or> content width
unix: set column width to header <or> content width
darwin: set column width to header

Apart from not setting to content width on macOS, it does not even set the width correctly to what is needed for the header. This seems to be a bug in macOS 12.1? Can anybody confirm? Great to have a non public bug tracker /s

I also found [sizeToFitWidthOfColumn](https://developer.apple.com/documentation/appkit/nstableviewdelegate/1526429-tableview?language=objc) in the Apple docs which does explain <sub>a little bit</sub> how Apple does their sampling.

szanni:
Not quite sure what is happening with the test runner on windows shared.

cody271:
> Apart from not setting to content width on macOS, it does not even set the width correctly to what is needed for the header. This seems to be a bug in macOS 12.1? Can anybody confirm? Great to have a non public bug tracker /s

Page16 test does correctly resize to the needed header width with `-1` here on macOS 11.

cody271:
Good news is that this works perfectly on `unix` and `windows`. 👍 

@szanni Happy to merge as is, or we can keep trying to fix the `darwin` issue. Up to you.

szanni:
> @szanni Happy to merge as is, or we can keep trying to fix the `darwin` issue. Up to you.

Yes, looks good to me. Got a TODO in the darwin code and a note in `ui.h` making users aware of slight differences across platforms.

There is still one inconsistency I'm noticing and that is the min width on darwin which is `10` on my system. This affects the `uiTableColumnSetWidth` function as it does manual user resizing. Should we set the min width to `0` on column creation as seems to be the case for unix and windows (windows seems to have no min width concept at all as far as I can tell)?

The reason I'm asking is because even setting min width and width to `0`, it still does not look like a `0` width column... so seems a little futile. I guess this is fine in a separate PR though? The code in `tablecolumn.m` really needs some love.

cody271:
> The reason I'm asking is because even setting min width and width to `0`, it still does not look like a `0` width column... so seems a little futile. I guess this is fine in a separate PR though? The code in `tablecolumn.m` really needs some love.
>
Futile for now indeed since it appears only `darwin` strictly enforces min width.

Only way to make the behavior totally consistent, requires messing with `NSTableView` internals inside our `uipriv` subclasses. (Which will then probably change from version to version of macOS.)

cody271 (approved):
LGTM

PR #44, opened 2022-01-20
[Old PR libui#516] Add new API function to set a table row double click callback.

szanni:
Old PR [libui#516](https://github.com/andlabs/libui/pull/516)

Proposal to add a new API function to set a callback for when a user double clicks a table row.
```c
void uiTableOnRowDoubleClicked(uiTable *t, void (*f)(uiTable *t, int row, void *data), void *data);
```

Implementations provided for darwin, unix, and windows. A small sample usage is provided in page16.

Only thing I'm not quite sure about is the naming. Alternatives would be `uiTableRowOnDoubleClicked` or instead of *DoubleClicked* use *Activated* if we should ever want to support single click activation and disable row selection?

### DIFF
--- a/darwin/table.h
+++ b/darwin/table.h
@@ -18,6 +18,8 @@ struct uiTable {
 	uiTableModel *m;
 	void (*headerOnClicked)(uiTable *, int, void *);
 	void *headerOnClickedData;
+	void (*onRowDoubleClicked)(uiTable *, int, void *);
+	void *onRowDoubleClickedData;
 };
 
 // tablecolumn.m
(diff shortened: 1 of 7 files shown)


cody271:
> Only thing I'm not quite sure about is the naming. Alternatives would be `uiTableRowOnDoubleClicked` or instead of _DoubleClicked_ use _Activated_ if we should ever want to support single click activation and disable row selection?
>
Yes `uiTable` should ideally support both row selection and activate on click with this API.

szanni:
> Yes `uiTable` should ideally support both row selection and activate on click with this API.

I will look into it as I'm fiddling with a selection API anyways. In theory you could spin your own with a `uiTableOnSelectionChanged` API, but activating on single click seems to be something that is common on mac and unix at least.

PR #47, opened 2022-01-24
[Linux+Windows] file/folder dialog parameters: path, name, filters

msink:
Parts from andlabs/libui#375 and andlabs/libui#190

Untested and Unfinished: macOS implementation missed

### DIFF
--- a/unix/stddialogs.c
+++ b/unix/stddialogs.c
@@ -6,23 +6,66 @@
 
 #define windowWindow(w) (GTK_WINDOW(uiControlHandle(uiControl(w))))
 
-static char *filedialog(GtkWindow *parent, GtkFileChooserAction mode, const gchar *confirm)
+static char *filedialog(GtkWindow *parent, GtkFileChooserAction mode, const gchar *confirm,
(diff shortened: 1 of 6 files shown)


unix/stddialogs.c
```diff
@@ -6,23 +6,66 @@
...
+	int i;
+	int s;
+
 	fcd = gtk_file_chooser_dialog_new(NULL, parent, mode,
 		"_Cancel", GTK_RESPONSE_CANCEL,
 		confirm, GTK_RESPONSE_ACCEPT,
 		NULL);
 	fc = GTK_FILE_CHOOSER(fcd);
+
+	for (i = s = 0; i < 255; i++) {
```
mohad12211, line 30: this breaks `uiOpenFolder` because `filter` is `NULL`
`for (i = s = 0; i < 255 && filter != NULL; i++) {` should fix it, right?

msink: or
```
if (!filter) filter = ""'
for (i = s = 0; i < 255; i++) {
```
entire loop here wants to be rewritten in more ideomatic way...

PR #54, opened 2022-01-25
[Linux] fix drawmatrix transform order to match Windows.

msink:
same as andlabs/libui#383

### DIFF
--- a/unix/drawmatrix.c
+++ b/unix/drawmatrix.c
@@ -31,35 +31,41 @@ static void c2m(cairo_matrix_t *c, uiDrawMatrix *m)
 void uiDrawMatrixTranslate(uiDrawMatrix *m, double x, double y)
 {
 	cairo_matrix_t c;
+	cairo_matrix_t tmp;
 
 	m2c(m, &c);
-	cairo_matrix_translate(&c, x, y);
+	cairo_matrix_init_translate(&tmp, x, y);
+	cairo_matrix_multiply(&c, &c, &tmp);
 	c2m(&c, m);
 }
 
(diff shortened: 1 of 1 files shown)


cody271:
I don't think any of the "Page 6" tests in`test/drawtests.c` expose this bug?
Does your logo [test](https://github.com/msink/kotlin-libui/blob/master/samples/logo/src/nativeMain/kotlin/main.kt) still expose this bug? Some screenshots of the different behavior would also help.

cody271:
`Direct2D: How to Scale an Object` from tester "Page 6" perhaps? I will try that.

cody271:
Okay just tested with "Page 6" and confirmed: Windows code is correct, other two are wrong.
I will need to port the fix over to macOS next.

See [old issue](https://github.com/andlabs/libui/issues/331)

cody271:
Update: this PR doesn't actually work..

Still wrong behavior on Linux, even with this patch applied:
drawmatrix_unix

Correct on Windows:
drawmatrix_windows

You can see that the smaller rectangle is supposed to be inside the larger one.

MSDN screenshot for reference:
scale-ovw

msink:
Well, it was just cherry-pick from https://github.com/Arisotura/melonDS/commit/dd529f0f5cdf37e4196c1793c3c538be8ebe4cd9
I personall  am completely ignorant in that tranformations, sorry.

Maybe it's jet another bug,

cody271:
Yeah, affine transformations can be a bit tricky.. I've now fixed the bug on Linux and macOS:

drawmatrix_unix_fixed
drawmatrix_darwin_fixed

cody271:
Replaced by #265.

Issue #31, opened 2022-01-16
uiImageView control


cody271:
[Old PR](https://github.com/andlabs/libui/pull/209)

AndyObtiva:
Thank you for continuing to maintain libui-ng. I recently worked on adding the newly implemented Table Selection API to [Glimmer DSL for LibUI](https://github.com/AndyObtiva/glimmer-dsl-libui) and it is almost ready for release.

Otherwise, we need this feature (image display support) very badly in [Glimmer DSL for LibUI](https://github.com/AndyObtiva/glimmer-dsl-libui). I've gotten multiple requests for it including this [new one](https://github.com/AndyObtiva/glimmer-dsl-libui/issues/42). It seems everyone who wants to build a desktop application will end up needing to render an image in their application at one point. Is there any ETA on when it will be available!?

By the way, the way [SWT](https://www.eclipse.org/swt/) handles this is by allowing us to display an image:
- inside a label (as a property on label)
- inside a canvas (like the libui area control)
- as application icon (shows up on ALT+TAB/CMD+TAB) 
- as a tab image-icon in tab folders.

cody271:
> It seems everyone who wants to build a desktop application will end up needing to render an image in their application at one point.
>
Yes absolutely, this is a sorely missing feature.

>  Is there any ETA for when it will be available!?
>
Basic `uiImageView` control, `uiImage` loading from file, and `uiArea` support are WIP.

> By the way, the way [SWT](https://www.eclipse.org/swt/) handles this is by allowing us to display an image:
> - as application icon (shows up on ALT+TAB/CMD+TAB)
>
App (and tray) icons are also a high priority. I'm not sure about the other image related feature requests listed.

petabyt:
IMO `uiImage` and `uiImageView` would be confusing for new users. (Shouldn't it be `uiImageControl`?)
Also appending a `uiImage` to a box would cause a segmentation fault. This already has caused some confusion (https://github.com/libui-ng/libui-ng/issues/229)

Why not just turn `uiImage` into a widget?
```
uiImage *img = uiNewImage(64, 64);
uiImageFromFile(img, "ide/libui.png");
uiBoxAppend(vbox, uiControl(img), 0);
```
This would keep backwards compatibility with the current code (the `uiImageFromFile` API is a suggestion, probably needs improvement)

AndyObtiva:
[...]

Issue #30, opened 2022-01-16
uiDrawBitmap


cody271:
[Old PR](https://github.com/andlabs/libui/pull/386)

`uiDrawBitmap` and `uiImage` are in fact different:

`uiImage` is an image _asset_ (immutable)
`uiDrawBitmap` is an image _buffer_ for raster drawing (mutable)

msink:
Well, if new class `uiDrawBitmap` is really buffer - maybe then just name it appropriately?
And leave `uiImage` as is.

variants:
- `uiImageBuffer`
- `uiImageSurface`

kojix2:
In the libui binding for the Ruby language, glimmer-dsl-libui provides the ability to display images in an Area. It reads a png image, extracts pixels, and arranges rectangles in the area. (Jaggy correction is not applied at this time) If uiDrawBitmap is completed, we may be able to replace the Ruby code with libui-ng's native functions.

matyalatte:
Sent a new PR with macOS support! #257

cody271:
>> If uiDrawBitmap is completed, we may be able to replace the Ruby code with libui-ng's native functions.
>>
>
> Sent a new PR with macOS support! #257
>
^ @kojix2

PR #55, opened 2022-01-25
[Linux+Windows] add uiImageBuffer APIs

msink:
adaptaion of andlabs/libui#386 + one more commit

- untested
- macOS missed

### DIFF
--- a/ui.h
+++ b/ui.h
@@ -525,7 +525,7 @@ _UI_EXTERN void uiDrawSave(uiDrawContext *c);
 _UI_EXTERN void uiDrawRestore(uiDrawContext *c);
 
 // ImageBuffer API
-_UI_EXTERN uiImageBuffer *uiNewImageBuffer(uiDrawContext *c, int width, int height);
+_UI_EXTERN uiImageBuffer *uiNewImageBuffer(uiDrawContext *c, int width, int height, int alpha);
 _UI_EXTERN void uiImageBufferUpdate(uiImageBuffer *buf, const void *data);
 _UI_EXTERN void uiImageBufferDraw(uiDrawContext *c, uiImageBuffer *buf, uiRect *srcrect, uiRect *dstrect, int filter);
 _UI_EXTERN void uiFreeImageBuffer(uiImageBuffer *buf);
--- a/ui.h
+++ b/ui.h
@@ -379,6 +379,8 @@ typedef struct uiDrawMatrix uiDrawMatrix;
 
 typedef struct uiDrawBrushGradientStop uiDrawBrushGradientStop;
 
+typedef struct uiImageBuffer uiImageBuffer;
+
 _UI_ENUM(uiDrawBrushType) {
 	uiDrawBrushTypeSolid,
 	uiDrawBrushTypeLinearGradient,
@@ -466,6 +468,15 @@ struct uiDrawStrokeParams {
 	double DashPhase;
 };
 
+struct uiRect {
+	int X;
+	int Y;
+	int Width;
+	int Height;
+};
+
+typedef struct uiRect uiRect;
+
 _UI_EXTERN uiDrawPath *uiDrawNewPath(uiDrawFillMode fillMode);
 _UI_EXTERN void uiDrawFreePath(uiDrawPath *p);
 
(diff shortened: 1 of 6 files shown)


mohad12211:
afaik, uiImage can only be used in a table, with this PR, can I show an image in a box for example?

msink:
This is not related to uiImage, and now I'm not sure that renaming it from uiDrawBitmap as was good idea, if someone so easy confuses there.
See #30

mohad12211:
I'm sorry, I just discovered this fork, and i'm still not familiar with the work that has been done here. the only thing that is holding me from using this library is that I can't show an image, uiImage is only used for table items, and I was asking about the progress done and wondered what does this PR does exactly?

msink:
> what uiImageBuffer (uiDrawBitmap) does exactly

Drawing a bitmap image on uiArea

msink:
And to your original question - yes, it can be used in a uiArea placed in a uiBox.

cody271:
> I'm sorry, I just discovered this fork, and i'm still not familiar with the work that has been done here.
>
Hi @mohad12211! For all the work that has been done so far, take a took at the [changelog](https://libui-ng.github.io/libui-ng/md__c_h_a_n_g_e_l_o_g.html).

> afaik, uiImage can only be used in a table, with this PR, can I show an image in a box for example?
>

I think you're probably looking for [`uiImageView`](https://github.com/libui-ng/libui-ng/issues/31) control, to show an image in a `uiBox`.

`uiDrawBitmap` is only for drawing images inside `uiArea` control, or for drawing new images at runtime with `uiDrawContext`.

mohad12211:
```c
static uiImageBuffer *img;
static unsigned char *binaryData;
static uiAreaHandler handler;

static void handlerDraw(uiAreaHandler *ah, uiArea *area, uiAreaDrawParams *p) {
  if (!binaryData)
    return;
  img = uiNewImageBuffer(p->Context, 320, 180, 1);
  uiRect src = {0, 0, 320, 180};
  uiRect dest = {0, 0, 320, 180};
  uiImageBufferUpdate(img, (void *)binaryData);
  uiImageBufferDraw(p->Context, img, &src, &dest, 0);
  uiFreeImageBuffer(img);
}
```
I used this to display an image, it does work with no problems (linux)
but it redraws the image very often, even when I just hover with the mouse over some buttons and stuff
is there a way to do this better or in a more optimized way?

cody271:
> 
> is there a way to do this better or in a more optimized way?
>
It looks like you are re-creating and then destroying `img` on each draw callback. Only create `img` object once, probably best to be at the same place in your code that you allocate `binaryData`. Then something like this should work fine:

```c
static void handlerDraw(uiAreaHandler *ah, uiArea *area, uiAreaDrawParams *p) {
  if (!img)
    return;
  uiImageBufferDraw(p->Context, img, &src, &dest, 0);
}
```

Also please try to provide screenshots when submitting bug reports if possible! Thanks

bcampbell:
I've got an annoying feature request :-)

The proposed `uiImageBuffer` kind of assumes you're happy to copy your image data into an already-allocated static image before drawing it to the screen.
But the ability I _really_ want is to be able to blit raw image data (i.e. allocated and managed by my own code) into a uiDrawContext. I have raw pixels that I want to get on the screen as fast as possible.

Ideally, I'd like a function like:
```
void uiBlat(uiDrawContext* ctx, int destx, int desty, const void* raw, int w, int h, ptrdiff_t stride, PixelFormat fmt);
    destx, desty - Coordinates on destination drawing area
    raw - A raw pointer to my source pixel data
    w,h - Dimensions of rectangle to copy
    stride - Distance between start of each row in memory (ideally should cope with -ve values to handle bottom-up images!)
    fmt - The byte ordering of the pixelformat (eg RGB, BGR, RGBx, ABGR...  whatever!).
            To support R,G,B, A (alpha), x (pad byte).
```
And that rectangle would just magically be blitted into the drawing area.
Bonus points for using the alpha for blending.
Extra bonus points for a full bitmask system to support arbitrary pixel formats (eg be able to support stuff like 565 16 bit RGB). (eg [SDL_PixelFormat](https://github.com/libsdl-org/SDL/blob/main/include/SDL_pixels.h#L170))

### Rationale:
My pixel paint/animation app, [EvilPixie](https://github.com/bcampbell/evilpixie) has a GUI-toolkit-agnostic core, so all the image management is handled by my own code. All I need from the GUI is a way to get chunks of raw image to the screen as fast as possible. I currently use Qt as my GUI toolkit, but I'd _love_ to switch to libUI(-ng)! But I can't with the currently-proposed uiImageBuffer.

### Prior art:
Qt QImage class is able to 'borrow' data:
https://doc.qt.io/qt-6/qimage.html#QImage-6

So the bit of code which paints my internal (non-gui-toolkit-specific) images to the screen is nice and simple and avoids extra copies:
https://github.com/bcampbell/evilpixie/blob/master/src/qt/editviewwidget.cpp#L86

cody271:
> 
> But the ability I _really_ want is to be able to blit raw image data (i.e. allocated and managed by my own code) into a uiDrawContext.
>

Yes thinking about images support some more here .. I am actually leaning towards adding [raw blitting](https://en.wikipedia.org/wiki/Bit_blit) support _first_, because this single new API call is such a simple and powerful primitive.

PR #56, opened 2022-01-27
[Old PR libui#497] Set column title on OS X 10.9.

szanni:
Old PR [libui#497](https://github.com/andlabs/libui/pull/497)

As can seen in the description of the original PR [the setTitle function is only available from macOS 10.10 onwards](https://developer.apple.com/documentation/appkit/nstablecolumn/1526875-title?language=objc).

The README states support for macOS 10.8.

I do not have a system that old, but I just tested the `[col headerCell] setStringValue:str]` branch and it works fine.

Only concert could be, why two branches, if the old way still works on newer systems?

### DIFF
--- a/darwin/tablecolumn.m
+++ b/darwin/tablecolumn.m
@@ -32,6 +32,15 @@
 	return nil;			// appease compiler
 }
 
+static void setTitle(uiprivTableColumn *col, NSString *str)
+{
+	if ([col respondsToSelector:@selector(setTitle:)]) {
+		[col setTitle:str];
+	} else {
+		[[col headerCell] setStringValue:str];
+	}
+}
+
 @end
 
 struct textColumnCreateParams {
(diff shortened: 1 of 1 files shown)


cody271:
> Only concert could be, why two branches, if the old way still works on newer systems?

We can't support release that old..

Mojave 10.14 or later. That's the oldest release to have received a security update within the past year.

(and will be Catalina 10.15 minimum soon.)

szanni:
> > Only concert could be, why two branches, if the old way still works on newer systems?
> 
> We can't support release that old..
> 
> Mojave 10.14 or later. That's the oldest release to have received a security update within the past year.
> 
> (and will be Catalina 10.15 minimum soon.)

Then we should reflect that in the README instead. Although I would be in favor of supporting older releases, as long as it does not become a maintenance burden...

I personally still run Windows XP for my Windows needs. VM, no network, works like a charm.

cody271:
> Although I would be in favor of supporting older releases, as long as it does not become a maintenance burden...

Yes its an absolute maintenance and testing (!) nightmare.

To be blunt: there's a reason nobody offers an LTS release of their software for free..

szanni:
> > Although I would be in favor of supporting older releases, as long as it does not become a maintenance burden...
> 
> Yes its an absolute maintenance and testing (!) nightmare.
> 
> To be blunt: there's a reason nobody offers an LTS release of their software for free..

I fully understand and sympathize. I guess this should be marked as WONTFIX then?

cody271:
> Then we should reflect that in the README instead.

Absolutely, the minimum versions specified there are not edited yet- still six years out of date.

PR #57, opened 2022-01-28
Test big tables

szanni:
As mentioned in another comment I believe we should ensure adequate performance on big tables.

To do so I increased the row count to 100k on page 16 & 17.

Testing things, unix and darwin performance are great and windows is ok (tested on a windows 7 vm).

Sadly wine performance is abysmal (deleting rows) but that is clearly an upstream issue that doesn't concern us. It just means I'll have to do more of my testing in a windows vm :|

### DIFF
--- a/test/page16.c
+++ b/test/page16.c
@@ -1,6 +1,8 @@
 // 21 june 2016
 #include "test.h"
 
+#define MODEL_NUM_ROWS 100000
+
 static uiTableModelHandler mh;
 
 static int modelNumColumns(uiTableModelHandler *mh, uiTableModel *m)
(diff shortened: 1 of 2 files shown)


cody271:
Ideally, this would be dynamic instead of hardcoded.

How difficult is it to only create the super large table when enabled with a new menu item/checkbox?

szanni:
> Ideally, this would be dynamic instead of hardcoded.

Good point. I've gotten so used to changing the hard coded values for local testing that I didn't even think of it.

> How difficult is it to only create the super large table when enabled with a new menu item/checkbox?

I would say that really depends. Once the table is published we can only grow it by using `uiTableModelRowInserted`.

So we could either have a button that just adds a few thousand rows at a time or create a new table _once_ entirely on demand.
At least that is what comes to my mind. From an implementation standpoint the first option is probably easiest. page17 can already add single rows.

cody271:
> So we could either have a button that just adds a few thousand rows at a time or create a new table _once_ entirely on demand.
> At least that is what comes to my mind. From an implementation standpoint the first option is probably easiest. page17 can already add single rows.

Adding a `Bulk Insert` button there is the way to go. 👍

PR #62, opened 2022-01-31
[Linux+Windows] add API for getting/setting window size

msink:


szanni:
I like the API in general. I added a few annotations, the biggest one probably being:
Do we want to support getting one dimension like `uiWindowPosition(w, NULL, &y);` like the unix code does?

And to make this useful we probably want a way to get the screen dimensions. Otherwise you can only really place the window in the top left corner?
That and a test case in the tester. I would love to have something more automated, but at least a small test/example in the tester for manual testing should IMO be included in all new PRs introducing new API calls?

mohad12211:
I have implemented the macos code, but an issue we need to discuss is wayland...
currently the only problem regarding wayland is borderless, although setting a window to borderless does work on gnome wayland, it doesn't work on kde plasma wayland. I might have to research this more...
this PR of course doesn't work at all on wayland. and the idea of wayland is to leave this to the wm/compositor. 
and I agree with this personally, the user's wm/compositor should have complete control over the window's geometry.
what do you all think?

cody271:
> what do you all think?

It is up to the individual compositor implementations as to whether they honor our windowing requests or not. From our perspective, nothing for us to do differently than any other GTK application..

From the official GTK docs for `gtk_window_move()`:
`Asks the [window manager][gtk-X11-arch] to move window to the given position. Window managers are free to ignore this`

I think borderless and fullscreen windows on Wayland are similar situation, we just make "best effort" requests.
(For example, the Phosh compositor _only_ allows borderless fullscreen windows, nothing else!)

szanni:
>  what do you all think?

We should document a warning about these functions just being hints on Unix. 
Are there MacOS or Windows systems out there that do the same? Apart from obscure things like Amethyst?

Happy to document these limitations in #96.

> It is up to the individual compositor implementations as to whether they honor our windowing requests or not. From our perspective, nothing for us to do differently than any other GTK application..

Correct. My system ignores all of these requests on X11, this has nothing to do with Wayland specifically. In X11 it depends on the windows manager. Most tiling window managers ignore positioning, window size and window decorations completely.

Issue #3, opened 2022-01-08
Generating doc from source code


inkydragon:
People want detailed documentation, but they don't want to write it themselves.
Automatically generating documentation from source code is a compromise.

We use C/C++, so maybe https://www.doxygen.nl/index.html is the best choice?
The following GUI frameworks also use doxygen:
- [wxWidgets](https://docs.wxwidgets.org/3.0/)
- [FLTK](https://www.fltk.org/doc-1.4/index.html)
- [JUCE](https://docs.juce.com/develop/index.html)

[Gtk](https://docs.gtk.org/gtk4/) uses their own written [gtk-doc](https://gitlab.gnome.org/GNOME/gtk-doc).

cody271:
You can take a look at my first attempt at this here: https://github.com/andlabs/libui/pull/175

inkydragon:
It looks like [jessevdk/cldoc](https://github.com/jessevdk/cldoc) has stopped developing.

cody271:
Hrmm yeah, cldoc looks to have been abandoned as well. :/

I guess to be fair, it has been nearly .. six years (2016) since I first tried to use cldoc in that old PR. A long time.

cody271:
> TODO:
> * [ ]  add example code to doc with some img. Maybe in another repo?
> 
Take a look at @parro-it's docs? Maybe we could just import [libui-node/docs](https://github.com/parro-it/libui-node/tree/master/docs) folder directly into [libui-ng/doc](https://github.com/libui-ng/libui-ng/tree/master/doc)?
(Also then would need to translate the JS example code back into plain C.)

cody271:
> TODO:
> * [ ]  ? Use a [modern theme](https://jothepro.github.io/doxygen-awesome-css/)
> 
Yes! 👍

cody271:
Also: Doxygen doesn't understand OOP style C automatically like cldoc.

We have to use special comments in `ui.h` like [this example](https://github.com/jserv/doxygen-oop-in-c/blob/master/manual.c).

inkydragon:
And anonymous enum in doxygen does not looks good:

Maybe apply a path:
```diff
+#ifdef __cplusplus
 #define _UI_ENUM(s) typedef unsigned int s; enum
+#else
+#define _UI_ENUM(s) enum s
+#endif
```

However, changes at the code level require some careful testing.
I will open a separate pr for this issue later.

cody271:
[...]

PR #64, opened 2022-02-03
annotate ui.h for doxygen

msink:
- use doxygen special commands for structuring
- missed documentatin copy-pasted from official Go bindings

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -389,7 +389,7 @@ IDL_PROPERTY_SUPPORT   = YES
 # all members of a group must be documented explicitly.
 # The default value is: NO.
 
-DISTRIBUTE_GROUP_DOC   = NO
+DISTRIBUTE_GROUP_DOC   = YES
 
 # If one adds a struct or class to a group and this option is enabled, then also
 # any nested class or struct is added to the same group. By default this option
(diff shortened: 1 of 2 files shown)


msink:
ready for review and hopefully merge

cody271:
> ready for review and hopefully merge

Much improved here, I think we're close on this! Remaining issues I see:

This is missing the [enum](https://github.com/libui-ng/libui-ng/issues/3#issuecomment-1013216508) annotations, and needs to restrict visibility of the private APIs that are in the header.
I think doxygen is also still a bit confused as to which of our C struct type defs are meant to be "classes", and which are just plain struct types.

Since @msink, @inkydragon, and myself all have a conflicting branch for this..
I'm going to combine all of our changes into one single PR now, with above issues resolved.

cody271:
Also- a few references to Go from the `andlabs/ui` godoc snuck in there .. 😛 

Need to keep everything plain C, will fix.

szanni:
Any progress on this? I would love to expand on the documentation in some areas when this hits upstream.

And as a side note: I personally find the classic
```c
/**
 *
 */
```
infinitely easier to parse, as there is less visual clutter. But maybe that's just because that is the style most projects use that I have come in contact with.

cody271:
Duplicate of #96

PR #69, opened 2022-02-05
darwin: Set minimum column width to 0 to match unix and windows.

szanni:
This is an attempt at more consistency between the platforms:

Unix and windows columns can manipulated by the user to have 0
width. The default on darwin is 10. This commit changes the minimum
width to 0 to match the other platforms.

Note: zero width on darwin still retains a significant amount of
space in the header compared to unix and windows. Still maybe better
than the default width of 10?

### DIFF
--- a/darwin/tablecolumn.m
+++ b/darwin/tablecolumn.m
@@ -597,6 +597,7 @@ void uiTableAppendTextColumn(uiTable *t, const char *name, int textModelColumn, 
 	col = [[uiprivTextImageCheckboxTableColumn alloc] initWithIdentifier:ident params:&p];
 	str = [NSString stringWithUTF8String:name];
 	[col setTitle:str];
+	[col setMinWidth: 0];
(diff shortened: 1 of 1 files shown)


cody271:
> Still maybe better than the default width of 10?
>
I can't discern any difference in behavior from 10 -> 0 here on macOS 11. What happens on macOS 12 ?

szanni:
> I can't discern any difference in behavior from 10 -> 0 here on macOS 11. What happens on macOS 12 ?

Interesting. I guess some defaults may have changed with macOS 12?
When I call `[tc minWidth]` on a table column I get a value of `10.0`.

Here are some screenshots highlighting the difference.

Without setting minWidth to 0 on macOS 12.2:
Screenshot 2022-02-06 at 17 20 19

With this PR on macOS 12.2
Screenshot 2022-02-06 at 17 19 42

PR #73, opened 2022-02-09
Table selection API

szanni:
This is an effort to provide a table selection API as was suggested in [libui#413](https://github.com/andlabs/libui/pull/413).

I have made some alterations to the original proposal.

This introduces 5 different table selection modes:
```
uiTableSelectionModeNone,       //!< Allow no row selection.                                
uiTableSelectionModeZeroOrOne,  //!< Allow zero or one row to be selected.                  
uiTableSelectionModeOne,        //!< Allow for exactly one row to be selected.              
uiTableSelectionModeZeroOrMany, //!< Allow zero or many (multiple) rows to be selected. 
```

With a get/set function:
```c
uiTableSelectionMode uiTableGetSelectionMode(uiTable *t);
void uiTableSetSelectionMode(uiTable *t, uiTableSelectionMode mode);
```

And then three functions to set a selection changed callback and get/free the actual selection:
```c
void uiTableOnSelectionChanged(uiTable *t, void (*f)(uiTable *t, void *data), void *data);
uiTableSelection* uiTableGetSelection(uiTable *t);
void uiTableSetSelection(uiTable *t, uiTableSelection *selection);
void uiFreeTableSelection(uiTableSelection* s);
```

uiTableSelection struct:
```c
struct uiTableSelection
{
    int NumRows;
    int *Rows;
};
```

I did some testing with the big tables (100k rows). Selecting all rows at once and having the tester print the rows on the terminal still has IMO very acceptable performance. Only minimal lag on my system.
An iterator style API would be nice in the future though. 

~~There is still one inconsistency across platforms remaining as far as I can tell:~~
~~Windows and unix both emit a selection changed when clicking on an already selected row on `uiTableSelectionModeZeroOrMany`, darwin will not. ~~Should be try to suppress this?~~ I've managed to suppress this for all other modes, for multi select I fear we might need to cache the entire selection and compare possibly on each selection change. Not sure what that would mean for performance.~~

~~All other edge cases _should_ be documented in the API docs.~~

Otherwise this works for me on all officially supported platforms. Wine sadly does not work, ~~see [upstream bug report](https://bugs.winehq.org/show_bug.cgi?id=52534). I've sent in patches for fixing wine, so far no luck.~~

I've kept my commits intact, if need be I can squash.

~~There are a few things I am not quite happy about:~~
* ~~Inefficiencies in the unix & darwin code. We need to iterate over the selection multiple times to convert formats~~
* ~~Should `uiTableSelectionCurrentSelection` return a value to indicate `malloc` errors? `assert` is not acceptable. Or is the current note in `ui.h` sufficient, requiring users to check BOTH `rows != NULL` and `numRows > 0`?~~

~~My question is: are these inefficiencies acceptable? I was wondering if it would be possible to provide some type of iterator API? I am however not sure if it is even possible or if it would really lead to any significant gains.~~

~~Then there seems to be a bug on windows (win7). Can anybody reproduce? Go to tester, page16, click `Multiple Selection`, try to edit a cell in the editable column. On many occasions the cell will loose focus by itself and editing will be stopped. Interestingly enough this does not happen on wine. Can anybody reproduce this on newer windows versions?~~

~~Edit: Windows default behavior when clicking an already selected/focused row seems to be: Send a deselect signal to for all items `nm->iItem == -1` and then send another select signal for the row that was already selected/focused. I replicated this with the [windows sample virtual list](https://github.com/szanni/Windows-classic-samples/tree/main/Samples/Win7Samples/winui/controls/common/vlistvw)~~

### DIFF
--- a/darwin/table.m
+++ b/darwin/table.m
@@ -8,6 +8,7 @@
 	uiTableModel *m;
 }
 - (id)initWithModel:(uiTableModel *)model;
+- (NSIndexSet *)tableView:(NSTableView *)tv selectionIndexesForProposedSelection:(NSIndexSet *)set;
 @end
 
 // TODO we really need to clean up the sharing of the table and model variables...
@@ -16,6 +17,8 @@
 	uiTableModel *uiprivM;
 	NSTableHeaderView *headerViewRef;
 }
+@property uiTableSelectionMode selectionMode;
+
 - (id)initWithFrame:(NSRect)r uiprivT:(uiTable *)t uiprivM:(uiTableModel *)m;
 - (uiTable *)uiTable;
 - (void)restoreHeaderView;
@@ -24,6 +27,8 @@
 @end
 
 @implementation uiprivTableView
+
+@synthesize selectionMode;
 
 - (id)initWithFrame:(NSRect)r uiprivT:(uiTable *)t uiprivM:(uiTableModel *)m
 {
@@ -104,6 +109,13 @@ static void setBackgroundColor(uiprivTableView *t, NSTableRowView *rv, NSInteger
 	return self;
 }
 
+- (NSIndexSet *)tableView:(uiprivTableView *)tv selectionIndexesForProposedSelection:(NSIndexSet *)set
+{
+	if ([tv selectionMode] == uiTableSelectionModeNone)
+		return nil;
+	return set;
+}
+
 - (NSInteger)numberOfRowsInTableView:(NSTableView *)tv
 {
 	return uiprivTableModelNumRows(self->m);
@@ -130,6 +142,12 @@ static void setBackgroundColor(uiprivTableView *t, NSTableRowView *rv, NSInteger
 {
 	uiTable *t = [tv uiTable];
 	t->headerOnClicked(t, [[tc identifier] intValue], t->headerOnClickedData);
+}
+
+- (void)tableViewSelectionDidChange:(NSNotification *)notification
+{
+	uiTable *t = [(uiprivTableView*)[notification object] uiTable];
+	t->onSelectionChanged(t, t->onSelectionChangedData);
 }
 
 @end
@@ -238,6 +256,52 @@ static void defaultHeaderOnClicked(uiTable *table, int column, void *data)
 	// do nothing
 }
 
+static void defaultOnSelectionChanged(uiTable *t, void *data)
+{
+	// do nothing
+}
+
+uiTableSelectionMode uiTableGetSelectionMode(uiTable *t)
+{
+	return [(uiprivTableView*)t->tv selectionMode];
+}
+
+void uiTableSetSelectionMode(uiTable *t, uiTableSelectionMode mode)
+{
+	switch (mode) {
+		case uiTableSelectionModeNone:
+			if ([t->tv numberOfSelectedRows] > 0)
+				[t->tv deselectAll: t->tv];
+			break;
+		case uiTableSelectionModeZeroOrOne:
+			if ([t->tv numberOfSelectedRows] > 1)
+				[t->tv deselectAll: t->tv];
+			[t->tv setAllowsMultipleSelection: NO];
+			[t->tv setAllowsEmptySelection: YES];
+			break;
+		case uiTableSelectionModeOne:
+			if ([t->tv numberOfSelectedRows] > 1)
+				[t->tv deselectAll: t->tv];
+			[t->tv setAllowsMultipleSelection: NO];
+			[t->tv setAllowsEmptySelection: NO];
+			break;
+		case uiTableSelectionModeZeroOrMany:
+			[t->tv setAllowsMultipleSelection: YES];
+			[t->tv setAllowsEmptySelection: YES];
+			break;
+		default:
+			uiprivUserBug("Invalid table selection mode %d", mode);
+			return;
+	}
+	[(uiprivTableView*)t->tv setSelectionMode: mode];
+}
+
+void uiTableOnSelectionChanged(uiTable *t, void (*f)(uiTable *, void *), void *data)
+{
+	t->onSelectionChanged = f;
+	t->onSelectionChangedData = data;
+}
+
 static void defaultOnRowClicked(uiTable *table, int row, void *data)
 {
 	// do nothing
@@ -260,6 +324,44 @@ void uiTableOnRowDoubleClicked(uiTable *t, void (*f)(uiTable *, int, void *), vo
 	t->onRowDoubleClickedData = data;
 }
 
+uiTableSelection* uiTableGetSelection(uiTable *t)
+{
+	__block int i = 0;
+	NSIndexSet *set = [t->tv selectedRowIndexes];
+	uiTableSelection *s = uiprivNew(uiTableSelection);
+
+	s->NumRows = [set count];
+	if (s->NumRows == 0)
+		s->Rows = NULL;
+	else
+		s->Rows = uiprivAlloc(s->NumRows * sizeof(*s->Rows), "uiTableSelection->Rows");
+
+	[set enumerateIndexesUsingBlock:^(NSUInteger row, BOOL *stop) {
+		s->Rows[i++] = row;
+	}];
+
+	return s;
+}
+
+void uiTableSetSelection(uiTable *t, uiTableSelection *sel)
+{
+	int i;
+	NSMutableIndexSet *set;
+	uiTableSelectionMode mode = [(uiprivTableView*)t->tv selectionMode];
+
+	if ((mode == uiTableSelectionModeNone && sel->NumRows > 0) ||
+	    (mode == uiTableSelectionModeZeroOrOne && sel->NumRows > 1) ||
+	    (mode == uiTableSelectionModeOne && sel->NumRows > 1)) {
+		// TODO log error
+		return;
+	}
+
+	set = [NSMutableIndexSet new];
+	for (i = 0; i < sel->NumRows; ++i)
+		[set addIndex: sel->Rows[i]];
+	[t->tv selectRowIndexes: set byExtendingSelection: FALSE];
+}
+
 uiTable *uiNewTable(uiTableParams *p)
 {
 	uiTable *t;
@@ -279,8 +381,6 @@ uiTable *uiNewTable(uiTableParams *p)
 	// TODO is this sufficient?
 	[t->tv setAllowsColumnReordering:NO];
 	[t->tv setAllowsColumnResizing:YES];
-	[t->tv setAllowsMultipleSelection:NO];
-	[t->tv setAllowsEmptySelection:YES];
 	[t->tv setAllowsColumnSelection:NO];
 	[t->tv setUsesAlternatingRowBackgroundColors:YES];
 	[t->tv setSelectionHighlightStyle:NSTableViewSelectionHighlightStyleRegular];
@@ -304,7 +404,9 @@ uiTable *uiNewTable(uiTableParams *p)
 	sp.VScroll = YES;
 	t->sv = uiprivMkScrollView(&sp, &(t->d));
 
+	uiTableSetSelectionMode(t, uiTableSelectionModeZeroOrOne);
 	uiTableHeaderOnClicked(t, defaultHeaderOnClicked, NULL);
+	uiTableOnSelectionChanged(t, defaultOnSelectionChanged, NULL);
 
 	// TODO WHY DOES THIS REMOVE ALL GRAPHICAL GLITCHES?
 	// I got the idea from http://jwilling.com/blog/optimized-nstableview-scrolling/ but that was on an unrelated problem I didn't seem to have (although I have small-ish tables to start with)
--- a/ui.h
+++ b/ui.h
@@ -3842,6 +3842,115 @@ _UI_EXTERN int uiTableColumnWidth(uiTable *t, int column);
  */
 _UI_EXTERN void uiTableColumnSetWidth(uiTable *t, int column, int width);
 
+/**
+ * Table selection modes.
+ *
+ * Table selection that enforce how a user can interact with a table.
+ *
+ * @warning An empty table selection is a valid state for any selection mode.
+ *          This is in fact the default upon table creation and can otherwise
+ *          triggered through operations such as row deletion.
+ *
+ * @enum uiTableSelectionMode
+ * @ingroup table
+ */
+_UI_ENUM(uiTableSelectionMode) {
+	/**
+	 * Allow no row selection.
+	 *
+	 * @warning This mode disables all editing of text columns. Buttons
+	 * and checkboxes keep working though.
+	 */
+        uiTableSelectionModeNone,
+        uiTableSelectionModeZeroOrOne,  //!< Allow zero or one row to be selected.
+        uiTableSelectionModeOne,        //!< Allow for exactly one row to be selected.
+        uiTableSelectionModeZeroOrMany, //!< Allow zero or many (multiple) rows to be selected.
+};
+
+/**
+ * Returns the table selection mode.
+ *
+ * @param t uiTable instance.
+ * @returns The table selection mode. [Default `uiTableSelectionModeZeroOrOne`]
+ *
+ * @memberof uiTable
+ */
+_UI_EXTERN uiTableSelectionMode uiTableGetSelectionMode(uiTable *t);
+
+/**
+ * Sets the table selection mode.
+ *
+ * @param t uiTable instance.
+ * @param mode Table selection mode to set.
+ *
+ * @warning All rows will be deselected if the existing selection is illegal
+ *          in the new selection mode.
+ * @memberof uiTable
+ */
+_UI_EXTERN void uiTableSetSelectionMode(uiTable *t, uiTableSelectionMode mode);
+
+/**
+ * Registers a callback for when the table selection changed.
+ *
+ * @param t uiTable instance.
+ * @param f Callback function.\n
+ *          @p sender Back reference to the instance that triggered the callback.\n
+ *          @p senderData User data registered with the sender instance.
+ * @param data User data to be passed to the callback.
+ *
+ * @note The callback is not triggered when calling uiTableSetSelection() or
+ *       when needing to clear the selection on uiTableSetSelectionMode().
+ * @note Only one callback can be registered at a time.
+ * @memberof uiTable
+ */
+_UI_EXTERN void uiTableOnSelectionChanged(uiTable *t, void (*f)(uiTable *t, void *data), void *data);
+
+/**
+ * Holds an array of selected row indices for a table.
+ *
+ * @struct uiTableSelection
+ * @ingroup table
+ */
+typedef struct uiTableSelection uiTableSelection;
+struct uiTableSelection
+{
+	int NumRows; //!< Number of selected rows.
+	int *Rows;   //!< Array containing selected row indices, NULL on empty selection.
+};
+
+/**
+ * Returns the current table selection.
+ *
+ * @param t uiTable instance.
+ * @returns The number of selected rows and corresponding row indices.\n
+ *          Data is owned by the caller, make sure to call uiFreeTableSelection().
+ *
+ * @note For empty selections the `Rows` pointer will be NULL.
+ * @memberof uiTable
+ */
+_UI_EXTERN uiTableSelection* uiTableGetSelection(uiTable *t);
+
+/**
+ * Sets the current table selection clearing any previous selection.
+ *
+ * @param t uiTable instance.
+ * @param sel Table selection.\n
+ *            Data is owned by the caller.
+ *
+ * @note Selecting more rows than the selection mode allows for results in nothing happening.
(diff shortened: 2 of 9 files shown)


szanni:
Some more debug info from the windows issue:

To be honest I have difficulty discerning what the code in `LVN_ITEMCHANGED:` really does.
The timers, the `!t->inLButtonDown && t->edit == NULL`, the nice TODO stating the code need cleanup???

What I have observed however is that when selecting a row everything behaves as it should. When I then try to edit one of the fields I get something like this:

```
# Click on row 9
LVN_ITEMCHANGED: 9
  Old
  New
    LVIS_FOCUSED
    LVIS_SELECTED
# Start editing field in row 9
# Then this starts happening
LVN_ITEMCHANGED: -1
  Old
    LVIS_SELECTED
  New
>> uiprivTableFinishEditingText()
LVN_ITEMCHANGED: 9
  Old
    LVIS_FOCUSED
  New
    LVIS_FOCUSED
    LVIS_SELECTED
>> !t->inLButtonDown && t->edit == NULL
```

Somehow the editing triggers windows to send out a signal that all rows `-1` have been de-selected.
Then it sends out another signal stating the previously selected row is now (re)selected.

I wonder if this is triggered due to some custom draw routines? wine does not exhibit this behavior.

Changing
```c
oldSelected = nm->uOldState & LVIS_SELECTED;
newSelected = nm->uNewState & LVIS_SELECTED;
```
To
```c
oldSelected = nm->uOldState & LVIS_FOCUSED;
newSelected = nm->uNewState & LVIS_FOCUSED;
```
Solves the premature editing stop but does not really help with the selection API, as we would still send out `uiTableSelectionOnChanged` twice although nothing has changed.

I would like a fix for windows sending out false `LVN_ITEMCHANGED` messages in the first place.

ui.h
```diff
@@ -1515,6 +1515,25 @@ _UI_EXTERN int uiTableColumnWidth(uiTable *t, int column);
 // Note: darwin currently only resizes to header title width on -1
 _UI_EXTERN void uiTableColumnSetWidth(uiTable *t, int column, int width);
 
+// uiTableSelectionAllowMultipleSelection() return if table allows for
+// multiple items being selected or not
+_UI_EXTERN int uiTableSelectionAllowMultipleSelection(uiTable *t);
```
bcampbell, line 1520: Should be called `uiTableAllowMultipleSelection()`? it operates on a `uiTable` rather than a `uiTableSelection` object.

ui.h
```diff
@@ -1515,6 +1515,25 @@ _UI_EXTERN int uiTableColumnWidth(uiTable *t, int column);
...
+
+// uiTableSelectionSetAllowMultipleSelection() set if table allows for
+// multiple items being selected or not
+_UI_EXTERN void uiTableSelectionSetAllowMultipleSelection(uiTable *t, int multipleSelection);
```
bcampbell, line 1524: should be `uiTableSetAllowMultipleSelection()`?

ui.h
```diff
@@ -1515,6 +1515,25 @@ _UI_EXTERN int uiTableColumnWidth(uiTable *t, int column);
...
+
+// uiTableSelectionOnChanged() set a callback function to be called when
+// the table selection changes
+_UI_EXTERN void uiTableSelectionOnChanged(uiTable *t, void (*f)(uiTable *t, void *data), void *data);
```
bcampbell, line 1528: maybe `uiTableOnSelectionChanged()` instead?
Rationale:
1. it reads better
2. seems more consistent with other APIs, eg `uiWindowOnContentSizeChanged()`

ui.h
```diff
@@ -1515,6 +1515,25 @@ _UI_EXTERN int uiTableColumnWidth(uiTable *t, int column);
...
+
+// uiTableSelectionCurrentSelection() get the current table selection
+// rows is a pointer to an array that will holds the current selection
+// numRows number of rows selected
+// Note: make sure to free() the array rows
+// Note: make sure to check if rows is !NULL even when numRows > 0 (indicates alloc error)
+_UI_EXTERN void uiTableSelectionCurrentSelection(uiTable *t, int* *rows, int *numRows);
```
bcampbell, line 1535: `uiTableCurrentSelection()`?

Also, I think it's a bad idea for the user to be `free()`-ing memory allocated inside libui-ng. (eg I could easily see libui-ng providing a mechanism where users could provide their own memory alloc/free functions).

Practically speaking, this means defining an object which is returned (eg `uiTableSelection`) with it's own deallocation API (eg `uiFreeTableSelection()`). All the other libui APIs work like this.
From memory, I think this is what I ended up doing in https://github.com/andlabs/libui/pull/413
(the struct was publicly defined, so you could read the selection data directly from it, and it even allowed user code to reorder the selected indices in place, but  reallocation was forbidden. You could instead go with an opaque struct, and provide an API for iteration, which might be better on some platforms maybe? not sure).

bcampbell (commented):
Cool - glad to see this!
I haven't tried it out properly, but here are my comments based on skimming through the diff.

ui.h
szanni: Oh wow. I'm slowly grasping the naming idea. My brain seems to have a hard time processing the camel case syntax.

ui.h
szanni: Interesting. This was actually something that bothered me in your original proposal. I personally dislike defining types for such use cases. Passing by pointer seems more C-like to me. This way I do not constantly have to look up the API documentation and can just use the languages native mechanisms. 
Digging through `ui.h` your approach looks more in line with the project though. Oh well.


> Also, I think it's a bad idea for the user to be `free()`-ing memory allocated inside libui-ng. (eg I could easily see libui-ng providing a mechanism where users could provide their own memory alloc/free functions).

I doubt providing a custom allocator would be possible. Hooking into macOS (objective-c)? Windows with C++? And the trickery called GLib? To my knowledge GTK is an absolute mess in that regard in general. They do not even free their own allocations if things haven't changed. A used to be a mess to track down what was happening when running applications through valgrind.

szanni:
Just added the remaining windows code.

Editing now does not get aborted anymore either.
The weird behavior that emits signals that all rows `nm->iItem == -1` have been deselected still persists when editing a cell. Not quite sure what to do with that. For those interested [here](https://github.com/szanni/libui-ng/tree/table-selection-debug) is a branch with my debug statements.

cody271:
> I updated the PR with the suggested changes by @bcampbell This should be pretty much complete now, if anybody would like to review.

Hi @bcampbell, have time to take another look for us by chance? Thanks!

bcampbell (approved):
All looks pretty sensible to me.

bcampbell:
One possible `uiTableSelection` optimization note:

We could save a malloc/free by allocating (and freeing) `uiTableSelection` and its row data in one continuous allocation eg:
```
uiTableSelection* sel = uiMallocOrWhateverItsCalled(sizeof(uiTableSelection) + sizeof(int)*numRows);
sel->NumRows = numRows;
sel->Rows = (int*)(sel+1);
...
```

szanni:
> All looks pretty sensible to me.

Perfect.

> One possible `uiTableSelection` optimization note:
> 
> We could save a malloc/free by allocating (and freeing) `uiTableSelection` and its row data in one continuous allocation eg:
> 
> ```
> uiTableSelection* sel = uiMallocOrWhateverItsCalled(sizeof(uiTableSelection) + sizeof(int)*numRows);
> sel->NumRows = numRows;
> sel->Rows = (int*)(sel+1);
> ...
> ```

I thought about something like that but I did not deem it worth the effort. It will just upset code checkers for IMO little benefit.

cody271 (changes requested):
Overall this looks great, API additions now totally consistent with the existing style.

I found only one minor issue remaining, in `uiTableCurrentSelection()`:
`s->Rows` can be NULL on `windows`, but `s->Rows` is always allocated even for empty selections on `unix`/`darwin`.
Behavior should probably match `windows` implementation on all platforms, allowing NULL for empty selections. And then documented as such in the struct definition.

szanni:
> `s->Rows` can be NULL on `windows`, but `s->Rows` is always allocated even for empty selections on `unix`/`darwin`.

Oh wow, `uiprivAlloc` looks completely broken. This is scary to look at. I was assuming it behaved like `malloc`, hence might return `NULL` on zero memory allocations. Now I understand why `uiprivFree` will error on `NULL`. I'll open an issue, as it completely disregards any alignment rules. It seems we are only getting lucky on ARM as buffers are accidentally aligned to `_Alignof(size_t)` and we are not allocating anything bigger than `size_t`.

>  allowing NULL for empty selections

I guess you were meaning to suggest _enforcing_. That was actually my initial intent. Good catch. I misremembered the C standard myself with regards to zero length `malloc`, which does not necessarily return `NULL`.

I'll fix that and document.

ui.h
```diff
@@ -1531,12 +1531,13 @@ _UI_EXTERN void uiTableOnSelectionChanged(uiTable *t, void (*f)(uiTable *t, void
 typedef struct uiTableSelection uiTableSelection;
 struct uiTableSelection
 {
-	int NumRows;
-	int *Rows;
+	int NumRows; //! Number of selected rows.
+	int *Rows; //! Array containing selected row indices, NULL on empty selection.
```
bcampbell, line 1535: How about:
```
int *Rows; //! Array containing selected row indices, undefined on empty selection.
```
I think the definitive is-selection-empty check should be `NumRows != 0`, and I don't think users should ever be null-checking the `Rows` pointer. Removing ambiguity is good.

(Besides, I can think of some future cases where you might potentially want a non-null Rows with an empty selection - eg shrinkable or truncatable selections. Unlikely, but seems silly to needlessly rule it out now).

szanni: Interesting. I might actually prefer the wording of undefined.
@cody271 what is your take?

I would love to actually support the other idea you had of one allocation only, but this can only be done in a standard compliant way via _flexible array members_, which are sadly only C99 and [NOT](https://en.wikipedia.org/wiki/Flexible_array_member#Availability) supported in C++.

cody271:
> Interesting. I might actually prefer the wording of undefined. @cody271 what is your take?
> 
>
For the sake of language bindings, we need to avoid any undefined pointers in the API as much as possible!
Either empty selections can be NULL, or guaranteed to always be an allocated pointer.

> (Besides, I can think of some future cases where you might potentially want a non-null Rows with an empty selection - eg shrinkable or truncatable selections. Unlikely, but seems silly to needlessly rule it out now).
>
If this in fact the case, then there is the answer: `Rows` must be guaranteed to always be an allocated pointer.
Otherwise NULL on empty still, if this future case can't be more well defined right now.

Regardless of which it is in the end, behavior needs to be the same on all platforms and documented as such.

bcampbell:
> For the sake of language bindings, we need to avoid any undefined pointers in the API as much as possible! Either empty selections can be NULL, or guaranteed to always be an allocated pointer.

By explicitly stating that Rows must be NULL for empty selections, we invite the user to rely on that for a empty-selection check when they really should just be checking `NumRows` instead. And then we're stuck with supporting that. So I'd like to avoid a Rows-is-NULL-for-empty-selections guarantee. Anything beyond that I don't really care too much about, because it can be changed in future without breaking anything.

(feel free to stop reading here - the rest is just pedantic detail!)

The only guarantee is that `Rows` can be indexed with any valid index up to `NumRows`.
So if `NumRows` is zero, you cannot index into `Rows` at all. So in that case it really doesn't matter if `Rows` points to valid allocated memory, NULL, or just some random number. From the callers (i.e. language bindings) point of view, it just doesn't matter what Rows is - you cannot legally deference it.

`uiTableCurrentSelection()` and `uiFreeTableSelection()` are the only things that need to know anything about the convention used for empty `Rows`.

This is assuming that language bindings use `uiTableCurrentSelection()` to create uiTableSelection objects, and `uiFreeTableSelection()` to free them, and don't reallocate Rows directly. Which seems like a reasonable restriction (i.e. language bindings shouldn't rely upon private implementation details inside libui-ng).

szanni:
> For the sake of language bindings, we need to avoid any undefined pointers in the API as much as possible! Either empty selections can be NULL, or guaranteed to always be an allocated pointer.

Can implementations not work around this? All the language FFIs seem to do fine with the C custom of undefined or implementation defined behavior. As stated, even `malloc()` is implementation defined on zero memory allocations.

> This is assuming that language bindings use `uiTableCurrentSelection()` to create uiTableSelection objects, and `uiFreeTableSelection()` to free them, and don't reallocate Rows directly. Which seems like a reasonable restriction (i.e. language bindings shouldn't rely upon private implementation details inside libui-ng).

I think this needs to be part of the documentation. All pointers returned by libui-ng are not to be reallocated or freed, but memory can otherwise be modified within their bounds. We might even want to expose `uiAlloc` or at least `uiRealloc` to the public for modifying memory?

I was thinking about programmatically selecting rows, not only modifying an existing selection. Something akin to `uiTableSetCurrentSelection()`. In that case it would indeed be annoying to enforce `Rows=NULL` when `NumRows==0`, as we would have to free the pointer and alloc when the selection changes. Maybe this should be part of this PR, as it might make requirements clearer? I haven't looked into this at all yet. Busy getting table selection fixes upstreamed into wine atm.

That being said - the lib is alpha stage. The entirety of the table API needs to be modified anyways. It is broken in so many places, I'm actually amazed it works at all. I'm still fiddling with automated testing as that will highlight how broken things truly are.

szanni:
Just added `uiTableSetCurrentSelection` to programmatically set table selections (tested on unix+windows, still waiting for my mac charger).

As stated in the API, Rows is never accessed on empty selections, so that does not aid in any way the prior discussion.

The other things I'm not quite happy about is the `uiprivUserBug` call, that essentially crashes the application if you do a multi select in single select mode. We either need a return value - or - a logging mechanism. Something like `uiLogError` or uiprivLogError.

szanni:
More API stuff decisions:
<hr>

What do we do on multiple selected and call to `uiTableSetAllowMultipleSelection(t, 0)`?

Behavior:
- unix: last selected item stays selected
- windows: everything stays selected
- darwin: ??? no test device

Ideas (descending in preference from my side):
1. log warning + copy unix behavior and keep last item selected
2. log warning + manually deselect all
3. keep it undefined

<hr>

What to do on `uiTableAllowMultipleSelection==0` and `uiTableSetCurrentSelection` with `NumRows > 1`?

Ideas:
1. log error + select last of said rows
2. log error + early return/do nothing
3. log error + deselect all

<hr>

I really want to get unit testing going so we can actually automatically test the behavior and make guarantees.

szanni:
Just updated with a new proposal. This ended up being way more complex than initially expected. The old proposal can still be found [here](https://github.com/szanni/libui-ng/tree/table-selection-v1).

Only seemingly still missing is suppressing `selectionChanged` for `uiTableSelectionModeZeroOrMany` when the selection has in fact not changed for unix and windows. I will still report the unix one upstream, but I'm not holding my breath for a gtk3 fix. Not sure what the performance penalty for a workaround is either.

Otherwise everything should be updated in the initial comment. I also tried implementing a `uiTableSelectionModeOneOrMany` mode, but that seems rather impossible for unix... at least I did not find an easy way.

szanni:
Found another inconsistency in ZeroOrOne and ZeroOrMany:
When deselecting the last row and then pressing up/down things behave differently on each platform.

Deselect the last row, press up/down:
- unix: the deselected row will get selected again
- windows: the next row (up/down) will get selected
- darwin: the first/last row of the table will get selected

Deselect the last row, press CMD/CTRL + up/down:
- unix: the focus moves (visible)
- windows: the focus moves (visible); for ZeroOrOne the next (up/down) row gets selected (that is only because I programmed it that way, easily changeable)
- darwin: the first/last row of the table gets selected

Do we care? Should we care? Selection works fine, the defaults of each platform just turn out to be different. Maybe this is actually desired, as this is what the user of said platform expects?

As there is no visual focus, and seemingly no way of selecting a row via keyboard on macOS, we can probably ignore that, as there is really nothing we can do about it.

What about unix/windows? I would probably vote for leaving the defaults alone. These are very much corner cases anyways, and things are perfectly function as they are.

Thought I'd bring this up anyways.

szanni:
The last commit should solve the double triggering of `onSelectioChanged` for windows. I know it looks like a lot of code.
The only other thing I can think of would be to cache the entire selection.

To test: on page16 the selection changed count should only go up on true selection changes. Try with CTRL/SHIFT/SPACE/CLICK/UP/DOWN/PAGEUP/PAGEDOWN/HOME/END, move and select via CTRL+SHIFT, then SHIFT+CLICK etc. I'm still compiling a full test sequence.

Missing:

- windows: Deleting and inserting rows
- unix: suppressing `onSelectionChanged` on multi select when the selection has not actually changed

szanni:
>     * unix: suppressing `onSelectionChanged` on multi select when the selection has not actually changed

Upstream [issue](https://gitlab.gnome.org/GNOME/gtk/-/issues/5314).

cody271:
> > ```
> > * unix: suppressing `onSelectionChanged` on multi select when the selection has not actually changed
> > ```
> 
> Upstream [issue](https://gitlab.gnome.org/GNOME/gtk/-/issues/5314).

In the short term, I think we'll have to cache the selection internally on Unix..

szanni:
> In the short term, I think we'll have to cache the selection internally on Unix..

Yeah, most likely unfortunately. I actually managed to patch the upstream code (will probably push tomorrow), but I sadly have little hope this will get merged.

szanni:
Nearly one year later and 36 commits squashed into one: I think I finally did it. Table selection!

Things should work as advertised. I updated the API proposal in the initial PR comment. All platform quirks should be fixed now and behavior should be the same on all.

If somebody wants to complain about `*Get*()` not being the ui way: look at `uiAttributeGetType` and `uiTableValueGetType`. It is already part of the API and IMO is the only sane way forward.

I hope I didn't miss any edge case. Behavior explained in the docs.

Review: @cody271 @bcampbell ? I know this is a lot of code. Sadly this is the insanity needed to make it work.

darwin/table.m
```diff
@@ -238,6 +256,49 @@ static void defaultHeaderOnClicked(uiTable *table, int column, void *data)
...
+		case uiTableSelectionModeOne:
+			if ([t->tv numberOfSelectedRows] > 1)
+				[t->tv deselectAll: t->tv];
+			[t->tv setAllowsMultipleSelection: NO];
+			[t->tv setAllowsEmptySelection: NO];
+			break;
+		case uiTableSelectionModeZeroOrMany:
+			[t->tv setAllowsMultipleSelection: YES];
+			[t->tv setAllowsEmptySelection: YES];
+			break;
```
cody271, line 291: Needs `default:` case at the end of the switch here to error out on invalid selection mode.

cody271:
> If somebody wants to complain about `*Get*()` not being the ui way: look at `uiAttributeGetType` and `uiTableValueGetType`. It is already part of the API where name collisions would occur and IMO is the only sane way forward.

Yes this is fine now, as we're much closer to make this breaking change across the whole API than previously.

cody271:
Should validate the selection mode given in each `uiTableSetSelectionMode()` impl, just like `uiTableSetSelection()` does [here](https://github.com/libui-ng/libui-ng/pull/73/files#diff-5c150d245024963a37d311e82d61efb023e6330a44e6648813811f96b3661a84R352). Otherwise looks good, latest `Page 16` tested and working!

darwin/table.m
szanni: Unsure what the default should do? Something like:
```
// TODO log error
return;
```
? Other suggestions?

cody271: Yes that would be fine for now, but we do also have `uiprivUserBug()`. Either okay until we have proper error logging framework in place.

windows/table.cpp
```diff
@@ -583,6 +770,55 @@ void uiTableHeaderSetVisible(uiTable *t, int visible)
...
+			t->maskOnSelectionChanged = TRUE;
+			ListView_SetItemState(t->hwnd, t->lastFocusedItem, LVIS_FOCUSED, LVIS_FOCUSED);
+			t->maskOnSelectionChanged = FALSE;
+			break;
+		}
+		// Fall through
+	case uiTableSelectionModeNone:
+		// Reset the table to it's initial state
+		ListView_SetItemState(t->hwnd, -1, 0, LVIS_SELECTED|LVIS_FOCUSED);
+		t->lastFocusedItem = -2;
```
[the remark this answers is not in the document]
szanni: So it is not `-1`. I'm unsure if this was just defensive programming or if this would trigger a signal (not) being emitted on some transition when changing modes.
This has to do with no early abort on deselect all (which sends) `nm->iItem = -1` for `uiTableSelectionModeZeroOrOne`.
We might otherwise trigger L458 or L463. Would need to see if that is possible.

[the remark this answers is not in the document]
cody271: Right since `-1` is valid for deselect all, we do need another special value for none. I would add a comment here.

darwin/table.m
szanni: Added a default case. Skipped my mind that we need to check as all enums get propagated to ints on export.

cody271 (approved):
LGTM

PR #75, opened 2022-02-17
Fix memory leak in filedialog.

mohad12211:
`gtk_file_chooser_get_filename` already returns `gchar*` that should be freed with `g_free()`, using `uiUnixStrdupText` will produce memory leak.

### DIFF
--- a/unix/stddialogs.c
+++ b/unix/stddialogs.c
@@ -28,7 +28,7 @@ static char *filedialog(GtkWindow *parent, GtkFileChooserAction mode, const gcha
 		gtk_widget_destroy(fcd);
 		return NULL;
 	}
-	filename = uiUnixStrdupText(gtk_file_chooser_get_filename(fc));
+	filename = gtk_file_chooser_get_filename(fc);
 	gtk_widget_destroy(fcd);
 	return filename;
 }


szanni:
Signed off.

cody271:
Is this same bug present with the other usages of `uiUnixStrdupText()`?

```shell
libui-ng$ grep -R uiUnixStrdupText unix
unix/group.c:	return uiUnixStrdupText(gtk_frame_get_label(g->frame));
unix/button.c:	return uiUnixStrdupText(gtk_button_get_label(b->button));
unix/label.c:	return uiUnixStrdupText(gtk_label_get_text(l->label));
unix/multilineentry.c:	// theoretically we could just return tret because uiUnixStrdupText() is just g_strdup(), but if that ever changes we can't, so let's do it this way to be safe
unix/multilineentry.c:	out = uiUnixStrdupText(tret);
unix/text.c:char *uiUnixStrdupText(const char *t)
unix/fontmatch.c:	uidesc->Family = uiUnixStrdupText(pango_font_description_get_family(pdesc));
unix/editablecombo.c:	out = uiUnixStrdupText(s);
unix/checkbox.c:	return uiUnixStrdupText(gtk_button_get_label(c->button));
unix/entry.c:	return uiUnixStrdupText(gtk_entry_get_text(e->entry));
unix/window.c:	return uiUnixStrdupText(gtk_window_get_title(w->window));
```

@mohad12211 Ideally we fix it everywhere in one PR, otherwise LGTM.

szanni:
Not quite:

> Is this same bug present with the other usages of `uiUnixStrdupText()`?
> 
> ```shell
> unix/group.c:	return uiUnixStrdupText(gtk_frame_get_label(g->frame));
> unix/button.c:	return uiUnixStrdupText(gtk_button_get_label(b->button));
> unix/label.c:	return uiUnixStrdupText(gtk_label_get_text(l->label));
>```
Docs state for all three `This string is owned by GTK and must not be modified or freed.`
>```shell
> unix/multilineentry.c:	// theoretically we could just return tret because uiUnixStrdupText() is just g_strdup(), but if that ever changes we can't, so let's do it this way to be safe
> unix/multilineentry.c:	out = uiUnixStrdupText(tret);
>```
Not a bug, just redundant copying, which gets freed though.
>```shell
> unix/text.c:char *uiUnixStrdupText(const char *t)
>```
-
>```
> unix/fontmatch.c:	uidesc->Family = uiUnixStrdupText(pango_font_description_get_family(pdesc));
>```
All good, from the docs `[...]and should not be freed.`
>```shell
> unix/editablecombo.c:	out = uiUnixStrdupText(s);
>```
Not a bug, just redundant copying.
>```shell
> unix/checkbox.c:	return uiUnixStrdupText(gtk_button_get_label(c->button));
> unix/entry.c:	return uiUnixStrdupText(gtk_entry_get_text(e->entry));
> unix/window.c:	return uiUnixStrdupText(gtk_window_get_title(w->window));
> ```
Again all look correct according to the GTK docs and need to be copied.

I would actually like to annotate `ui.h` with a `@note` whenever the caller need to free the memory when the doxygen switch happens.

msink:
AFAIK strings passed from language bindings can be freed/moved at any time after calling libui, in many cases - immediately after calling.
So if libui wants permanent string - it should be copied before passing to operating system.

szanni:
> AFAIK strings passed from language bindings can be freed/moved at any time after calling libui, in many cases - immediately after calling. So if libui wants permanent string - it should be copied before passing to operating system.

Not quite sure how to parse your comment.

All strings returned by libui need to be freed with `uiFreeText`. They can not be reallocated. I guess modifying characters would be fine, as the caller owns the buffer (with exception of realloc). If we would want to allow realloc we would have to change quite a few things around?

msink:
Well, for example binding language can use UTF16 encoding, while OS use UTF8.
So all strings passed to libui are converted in temporary buffer, and freed immediately after calling.

szanni:
The comment in `unix/multilineentry.c` highlights something interesting:
> // theoretically we could just return tret because uiUnixStrdupText() is just g_strdup(), but if that ever changes we can't, so let's do it this way to be safe

I personally believe future proofing in this way is just clutter, hence #81.
If we would want to plan for said imaginary future however - this PR would be incorrect - and should instead copy the behavior outlined in `unix/multilineentry.c`.

mohad12211:
I mean we can do the same as in `unix/multilineentry.c` for the sake of consistency aswell, but I agree with @szanni and his PR, prefer to keep it simple unless we implement such feature.

cody271 (approved):
LGTM

PR #81, opened 2022-04-30
Remove redundant uiUnixStrdupText calls.

szanni:
As was highlighted in #75 there might be some more nonsensical use of `uiUnixStrdupText()`.

This PR removes the future proofing for the imaginary future when the implementation of `uiUnixStrdupText()` would change.

I am honestly not quite sure what the original intent was, why not use `g_strdup` directly? Going through the commit logs, I could not find any explanation.

### DIFF
--- a/unix/editablecombo.c
+++ b/unix/editablecombo.c
@@ -33,14 +33,7 @@ void uiEditableComboboxAppend(uiEditableCombobox *c, const char *text)
 
 char *uiEditableComboboxText(uiEditableCombobox *c)
 {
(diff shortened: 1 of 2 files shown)


cody271:
> I am honestly not quite sure what the original intent was, why not use `g_strdup` directly? Going through the commit logs, I could not find any explanation.
>
The original intent was being able to add in allocations debugging in a single place in the code.

Nonsensical clutter at this point though..

cody271 (approved):
LGTM

Issue #85, opened 2022-05-06
Const correctness.


szanni:
There are various places within the API/code that should be marked as `const`.
TODO/TASK: find those

First of all, this is better style. Second of all, it signals more clearly to API users what to expect, especially for language bindings with more strict `const` enforcement than C.

dwRchyngqxs:
It also helps automatic ffi generators use the correct types. I have rust bindgen in mind.

Issue #84, opened 2022-05-06
Compile time warnings on all platforms.


szanni:
There are various compile time warnings on all platforms.
I believe it to be good practice to maintain a custom of zero warnings, as warnings are often a sign of design flaws or hidden errors.

TODO:
* [ ] unix
* [ ] darwin
* [ ] windows
* [ ] common
* [ ] update CONTRIBUTING.md to reflect a zero warning rule for new patches

We might want to enable `-Werror` too. The Linux kernel has just done so - and while it obviously can be very painful - it clearly works.
The other idea would be to compile with `-Werror` in the CI runners.

msink:
fixed remaining windows-mingw warnings in #83

PR #86, opened 2022-05-06
Make uiImageAppend const correct.

szanni:
Change API from
```c
uiImageAppend(uiImage *i, void *pixels, int pixelWidth, int pixelHeight, int byteStride);
```
to
```c
void uiImageAppend(uiImage *i, const void *pixels, int pixelWidth, int pixelHeight, int byteStride);
```
and fix the test case to make the compiler happy.

One thing to possibly do is change the type signature for `pixels` entirely.
Make it `const uint8_t *pixels` instead, not sure if there was a reason not to do so in the first place, as things get casted back to `uint8_t` in all implementations anyways.
This would make alignment rules very clear - especially for architectures with stricter alignment rules than `x86_64`.

Related to #84 #85


mohad12211:
There's also `-Wempty-body` warning at 
https://github.com/libui-ng/libui-ng/blob/baae3e47032ab64f936e6d8cecfe7df53771ab4f/unix/image.c#L44-L45
which can be fixed with
```c
	if (cairo_surface_status(cs) != CAIRO_STATUS_SUCCESS)
		{/* TODO */}
```

szanni:
> There's also `-Wempty-body` warning at
> 
> https://github.com/libui-ng/libui-ng/blob/baae3e47032ab64f936e6d8cecfe7df53771ab4f/unix/image.c#L44-L45
> 
> which can be fixed with
> 
> ```c
> 	if (cairo_surface_status(cs) != CAIRO_STATUS_SUCCESS)
> 		{/* TODO */}
> ```

See #87

PR #87, opened 2022-05-06
unix: Fix some of the easy to fix compile time warnings.

szanni:
This is a first step towards removing all compile time warnings, see #84.

This PR addresses easy to fix issues. More work is still needed - those patches might need some discussion/more thorough review though, hence separate PRs.

### DIFF
--- a/common/attribute.c
+++ b/common/attribute.c
@@ -62,7 +62,8 @@ static void destroy(uiAttribute *a)
 void uiprivAttributeRelease(uiAttribute *a)
 {
 	if (a->ownedByUser)
-		/* TODO implementation bug: we can't release an attribute we don't own */;
+		uiprivImplBug("Can't release attribute we don't own %p", a);
+
 	a->refcount--;
 	if (a->refcount == 0)
 		destroy(a);
@@ -71,7 +72,8 @@ void uiprivAttributeRelease(uiAttribute *a)
 void uiFreeAttribute(uiAttribute *a)
 {
 	if (!a->ownedByUser)
-		/* TODO user bug: you can't free an attribute you don't own */;
+		uiprivImplBug("Can't release attribute we don't own %p", a);
+
 	destroy(a);
 }
 
--- a/test/main.c
+++ b/test/main.c
@@ -46,8 +46,8 @@ int main(int argc, char *argv[])
 	const char *err;
 	uiWindow *w;
 	uiBox *page2, *page3, *page4, *page5;
-	uiBox *page6, *page7, *page8, *page9, *page10;
-	uiBox *page11, *page12, *page13;
+	uiBox *page6, *page7;
+	uiBox *page12, *page13;
 	uiTab *page14;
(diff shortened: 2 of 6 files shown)


mohad12211:
fixing `unix/text.c:16:16 -Wimplicit-function-declaration` also seems easy, we just add `#include <strings.h>` to `unix/uipriv_unix.h` unless i'm missing something

szanni:
> fixing `unix/text.c:16:16 -Wimplicit-function-declaration` also seems easy, we just add `#include <strings.h>` to `unix/uipriv_unix.h` unless i'm missing something Edit: see https://man7.org/linux/man-pages/man3/strcasecmp.3.html#NOTES

That would be the easy fix. The warning highlights a bigger issue for me though: why does `uiprivStricmp` exist in the first place?
IMO further research is needed. Not sure how `strcasecmp` behaves on windows, locale specific stuff, that's my I would put it in a separate PR + discussion. A lot of the `uipriv` functions seem to exist _for the sake of_. In quite a few instances I can not discern why the standard lib functions should not suffice.

cody271:
> That would be the easy fix. The warning highlights a bigger issue for me though: why does `uiprivStricmp` exist in the first place? IMO further research is needed. Not sure how `strcasecmp` behaves on windows, locale specific stuff, that's my I would put it in a separate PR + discussion. A lot of the `uipriv` functions seem to exist _for the sake of_. In quite a few instances I can not discern why the standard lib functions should not suffice.
>
I can provide a bit of the history here: the original purpose of the `uipriv` functions was mainly to help provide consistent behavior on Windows. Particularly for UTF-8 in strings. They were then also useful for being able to add in allocations debugging in a single place in the code.

That being said, these private function wrappers were taken a bit too far in a few places.. like with `uiUnixStrdupText()`. Really the most important thing here, in terms of the API, is just that strings are always freed with `uiFreeText()`.

szanni:
> I can provide a bit of the history here: the original purpose of the `uipriv` functions was mainly to help provide consistent behavior on Windows. Particularly for UTF-8 in strings. They were then also useful for being able to add in allocations debugging in a single place in the code.

While I never understood the allocation wrappers for debugging purposes... debuggers anybody?

It seems like the windows implementation is the only one that may actually work? `strcasecmp` used on darwin and unix will not do what you expect it to when using UTF-8 strings. Unless things have changed we would need a proper unicode library for that.

> That being said, these private function wrappers were taken a bit too far in a few places.. like with `uiUnixStrdupText()`. Really the most important thing here, in terms of the API, is just that strings are always freed with `uiFreeText()`.

The `uiFreeText()` makes a lot of sense, but yeah there seems to be a lot of optimization for _what ifs_. `Premature optimization is the root of all evil` comes to mind for me.
I think we should ensure all strings being returned by `ui` allow for normal string operations, essentially everything apart from `realloc` and `free`.

cody271 (approved):
LGTM

PR #93, opened 2022-05-17
[All Platforms] uiWindow: add event uiWindowOnFocusChanged() / uiWindowFocused()

mohad12211:
Tested on mac and linux. ~~Windows is still untested.~~
completes #52 
what do you think about the names GetFocus and LoseFocus? andlabs said [here](https://github.com/andlabs/libui/pull/378#issuecomment-411694239): 
> Not too happy with the names GetFocus/LoseFocus, as those names tend to refer to keyboard focus, not becoming the active window

should we change them? 
Edit: Tested it on Windows, does seem to work with no issues.

### DIFF
--- a/test/page5.c
+++ b/test/page5.c
@@ -68,12 +68,12 @@ static void msgBoxError(uiButton *b, void *data)
 
 void onGetFocus(uiWindow *w, void *data)
 {
-	uiCheckboxSetChecked(uiCheckbox(data), 1);
+	uiLabelSetText(uiLabel(data), "Window is focused");
 }
 
 void onLoseFocus(uiWindow *w, void *data)
 {
-	uiCheckboxSetChecked(uiCheckbox(data), 0);
+	uiLabelSetText(uiLabel(data), "Window is not focused");
 }
 
 
@@ -84,7 +84,7 @@ uiBox *makePage5(uiWindow *pw)
 	uiBox *hbox;
 	uiButton *button;
 	uiLabel *label;
-	uiCheckbox *check;
+	uiLabel *focusLabel;
 
 	parent = pw;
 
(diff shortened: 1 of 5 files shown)


cody271:
> should we change them?

I'm okay with the names `uiWindowOnGetFocus()` / `uiWindowOnLoseFocus()` as is here.

Anything related to keyboard focus could be prefixed with `Key` in the name for added clarity.

test/page5.c
```diff
@@ -66,12 +66,25 @@ static void msgBoxError(uiButton *b, void *data)
...
+
+
+
 uiBox *makePage5(uiWindow *pw)
 {
 	uiBox *page5;
 	uiBox *hbox;
 	uiButton *button;
 	uiLabel *label;
+	uiCheckbox *check;
```
cody271, line 87: The API is a read only event: you can't change the window focus programmatically.

Probably should be a label instead of a checkbox?

szanni:
> > Not too happy with the names GetFocus/LoseFocus, as those names tend to refer to keyboard focus, not becoming the active window

The typical ui convention would most likely be something like:
```c
void uiWindowOnFocusChanged(uiWindow *w, void (*f)(uiWindow *, void *), void *data);
int uiWindowFocused(uiWindow *w);
```
Or something like `uiWindowHasFocus`. The important part being one function signaling a change, and another function to be called by the user to determine the state. This way the state can be queried independently of any change.

cody271 (changes requested):
Needs to handle both the change event and querying the state:
```c
void uiWindowOnFocusChanged(uiWindow *w, void (*f)(uiWindow *, void *), void *data);
int uiWindowFocused(uiWindow *w);
```

cody271 (approved):
Page5 test is confirmed working. LGTM.

PR #94, opened 2022-05-18
Update ArchLinux AUR URL.

szanni:
Point to libui-ng instead of libui.


cody271:
> Point to libui-ng instead of libui.

And the new package supports ARM! 👍

cody271 (approved):
LGTM

PR #95, opened 2022-05-19
Fix uiWindowSetResizeable for unix

mohad12211:
On mac and windows, the behavior when calling `uiWindowSetResizeable(w, 0);` is that the current size is locked. but on unix, it resets to the smallest size which makes no sense, this should fix it, and from my tests, it doesn't break anything.

### DIFF
--- a/unix/window.c
+++ b/unix/window.c
@@ -237,10 +237,17 @@ int uiWindowResizeable(uiWindow *w)
 
 void uiWindowSetResizeable(uiWindow *w, int resizeable)
 {
-	int x, y;
+	// workaround for https://gitlab.gnome.org/GNOME/gtk/-/issues/4945
+	// calling gtk_window_set_resizable(w->window, 0) will cause the window to resize to default size
+	// (default is smallest size here because we're using gtk_window_resize() when creating the window)
+	// to prevent this we call gtk_window_set_default_size() on the current window size so that it doesn't resize
+	if (resizeable == 0) {
+		gint width, height;
+		gtk_window_get_size(w->window, &width, &height);
+		gtk_window_set_default_size(w->window, width, height);
+	}
+
 	w->resizeable = resizeable;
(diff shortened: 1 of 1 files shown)


szanni:
First of all, thanks for highlighting inconsistencies!!

I have not tested this yet (got to install a non tiling WM).
Did you find an upstream bug report?
This is an upstream bug to me, either in the implementation or the docs. Even if we merge this, this should IMO be reported upstream. I found this 10 year old stackoverflow [thread](https://stackoverflow.com/questions/8193613/gtk-window-resize-disable-without-going-back-to-default), they seemingly did not find a solution.

One thing I would request right off the bat with this patch though is a comment to _WHY_ we need this convoluted shenanigans. And possibly only do it on `if(resizable == 0)`.
And did you check if the callback `uiWindowOnContentSizeChanged` gets triggered by this? We would need to mask that signal as well.

mohad12211:
made the workaround more clean and documented the reason.
here's the upstream [issue](https://gitlab.gnome.org/GNOME/gtk/-/issues/4945).
they don't want to fix it for gtk3 as gtk4 doesn't have the problem...
also I checked that `uiWindowOnContentSizeChanged` is not triggered by this.

szanni (approved):
Tested. Works, fixes the issue and does not introduce any regressions as far as I can tell.

szanni:
> made the workaround more clean and documented the reason. here's the upstream [issue](https://gitlab.gnome.org/GNOME/gtk/-/issues/4945). they don't want to fix it for gtk3 as gtk4 doesn't have the problem... also I checked that `uiWindowOnContentSizeChanged` is not triggered by this.

Thanks for the upstream report. Sad to see it was tagged _wontfix_. Patch looks good. The first two lines of the comment would have been enough for me, but :)

cody271 (approved):
LGTM

Issue #92, opened 2022-05-15
Document default configuration of controls


norepimorphism:
Is `uiWindow` margined or unmargined by default? Is it bordered by default? I can find the answers to these questions through either trial-and-error or inspecting the source code, but I'd like this to be documented.

I think it's important to have clearly-defined defaults so that users of *libui-ng* can rely on default behavior instead of explicitly calling `uiWindowSetMargined` and `uiWindowSetBordered`, for example, on every `uiWindow` they create. It would also be nice if the default configurations were consistent across platforms, but if that's undesired, it should at least be documented.

I don't mind taking this issue on as I'm already exploring this for my Rust bindings.

szanni:
> Is `uiWindow` margined or unmargined by default? Is it bordered by default? I can find the answers to these questions through either trial-and-error or inspecting the source code, but I'd like this to be documented.
> 
> I think it's important to have clearly-defined defaults so that users of _libui-ng_ can rely on default behavior instead of explicitly calling `uiWindowSetMargined` and `uiWindowSetBordered`, for example, on every `uiWindow` they create. It would also be nice if the default configurations were consistent across platforms, but if that's undesired, it should at least be documented.

Fully agree. Clear defaults are important and I believe we should try to push for a default across platforms, where possible.
I already worked on some personal itches with `uiSlider` in the past.

I incorporated your idea in #96, but as you said, it is fairly trial and error right now - so I marked many as `TODO` for the time being.

PR #96, opened 2022-05-21
Doxygen documentation.

szanni:
I know there have been numerous attempts at annotating the library, here is another one.

Instead of dumping everything into one commit, I decided to go by component.

Documented (missing proper callback formatting)
- [x] uiWindow
- [x] uiButton
- [x] uiBox
- [x] uiCheckbox
- [x] uiEntry
- [x] uiLabel
- [x] uiTab
- [x] uiGroup
- [x] uiSpinbox
- [x] uiSlider
- [x] uiProgressBar
- [x] uiSeparator
- [x] uiCombobox
- [x] uiEditableCombobox
- [x] uiRadioButtons
- [x] uiDateTimePicker
- [x] uiMultilineEntry
- [x] uiOpenFile() uiOpenFolder() uiSaveFile() uiMsgBox() uiMsgBoxError()
- [x] uiImage
- [x] uiGrid
- [x] uiColorButton
- [x] uiForm
- [x] uiFontButton
- [x] uiMenuItem
- [x] uiMenu
- [x] uiTable & friends

### TODO

- [x] Annotate all `uiWindow` functions that are merely hints on unix as such
- [x] Document callbacks

Supercedes #64
Related to #3 #92

### DIFF
--- a/ui.h
+++ b/ui.h
@@ -74,8 +74,13 @@ _UI_EXTERN void uiOnShouldQuit(int (*f)(void *data), void *data);
 
 _UI_EXTERN void uiFreeText(char *text);
 
-typedef struct uiControl uiControl;
 
+/**
+ * Base class for GUI controls providing common methods.
+ *
+ * @struct uiControl
+ */
+typedef struct uiControl uiControl;
 struct uiControl {
 	uint32_t Signature;
 	uint32_t OSSignature;
@@ -94,24 +99,146 @@ struct uiControl {
 };
 // TOOD add argument names to all arguments
 #define uiControl(this) ((uiControl *) (this))
-_UI_EXTERN void uiControlDestroy(uiControl *);
-_UI_EXTERN uintptr_t uiControlHandle(uiControl *);
-_UI_EXTERN uiControl *uiControlParent(uiControl *);
-_UI_EXTERN void uiControlSetParent(uiControl *, uiControl *);
-_UI_EXTERN int uiControlToplevel(uiControl *);
-_UI_EXTERN int uiControlVisible(uiControl *);
-_UI_EXTERN void uiControlShow(uiControl *);
-_UI_EXTERN void uiControlHide(uiControl *);
-_UI_EXTERN int uiControlEnabled(uiControl *);
-_UI_EXTERN void uiControlEnable(uiControl *);
-_UI_EXTERN void uiControlDisable(uiControl *);
 
+/**
+ * Dispose and free all allocated resources.
+ *
+ * @param c uiControl instance.
+ * @memberof uiControl
+ */
+_UI_EXTERN void uiControlDestroy(uiControl *c);
+
+/**
+ * Returns the control's OS-level handle.
+ *
+ * @param c uiControl instance.
+ * @returns OS-level handle.
+ * @memberof uiControl
+ */
+_UI_EXTERN uintptr_t uiControlHandle(uiControl *c);
+
+/**
+ * Returns the parent control.
+ *
+ * @param c uiControl instance.
+ * @returns The parent control, `NULL` if detached.
+ * @memberof uiControl
+ */
+_UI_EXTERN uiControl *uiControlParent(uiControl *c);
+
+/**
+ * Sets the control's parent.
+ *
+ * @param c uiControl instance.
+ * @param parent The parent control, `NULL` to detach.
+ * @memberof uiControl
+ */
+_UI_EXTERN void uiControlSetParent(uiControl *c, uiControl *parent);
+
+/**
+ * Returns whether or not the control is a top level control.
+ *
+ * @param c uiControl instance.
+ * @returns `TRUE` if top level control, `FALSE` otherwise.
+ * @memberof uiControl
+ */
+_UI_EXTERN int uiControlToplevel(uiControl *c);
+
+/**
+ * Returns whether or not the control is visible.
+ *
+ * @param c uiControl instance.
+ * @returns `TRUE` if visible, `FALSE` otherwise.
+ * @memberof uiControl
+ */
+_UI_EXTERN int uiControlVisible(uiControl *c);
+
+/**
+ * Shows the control.
+ *
+ * @param c uiControl instance.
+ * @memberof uiControl
+ */
+_UI_EXTERN void uiControlShow(uiControl *c);
+
+/**
+ * Hides the control.
+ *
+ * @param c uiControl instance.
+ * @note Hidden controls do not take up space within the layout.
+ * @memberof uiControl
+ */
+_UI_EXTERN void uiControlHide(uiControl *c);
+
+/**
+ * Returns whether or not the control is enabled.
+ *
+ * Defaults to `true`.
+ *
+ * @param c uiControl instance.
+ * @see uiControlEnabledToUser
+ * @memberof uiControl
+ */
+_UI_EXTERN int uiControlEnabled(uiControl *c);
+
+/**
+ * Enables the control.
+ *
+ * @param c uiControl instance.
+ * @memberof uiControl
+ */
+_UI_EXTERN void uiControlEnable(uiControl *c);
+
+/**
+ * Disables the control.
+ *
+ * @param c uiControl instance.
+ * @memberof uiControl
+ */
+_UI_EXTERN void uiControlDisable(uiControl *c);
+
+/**
+ * Allocates a uiControl.
+ *
+ * Helper to allocate new controls.
+ *
+ * @param n Size of type to allocate.
+ * @todo Document parameters
+ * @memberof uiControl
+ */
 _UI_EXTERN uiControl *uiAllocControl(size_t n, uint32_t OSsig, uint32_t typesig, const char *typenamestr);
-_UI_EXTERN void uiFreeControl(uiControl *);
 
-// TODO make sure all controls have these
-_UI_EXTERN void uiControlVerifySetParent(uiControl *, uiControl *);
-_UI_EXTERN int uiControlEnabledToUser(uiControl *);
+/**
+ * Frees the control.
+ *
+ * @param c uiControl instance.
+ * @memberof uiControl
+ */
+_UI_EXTERN void uiFreeControl(uiControl *c);
+
+/**
+ * Makes sure the control's parent can be set to @p parent.
+ *
+ * @param c uiControl instance.
+ * @param parent uiControl instance.
+ * @todo Make sure all controls have these
+ * @warning This will crash the application if `FALSE`.
+ * @memberof uiControl
+ */
+_UI_EXTERN void uiControlVerifySetParent(uiControl *c, uiControl *parent);
+
+/**
+ * Returns whether or not the control can be interacted with by the user.
+ *
+ * Checks if the control and all it's parents are enabled to make sure it can
+ * be interacted with by the user.
+ *
+ * @param c uiControl instance.
+ * @returns `TRUE` if enabled, `FALSE` otherwise.
+ * @see uiControlEnabled
+ * @memberof uiControl
+ */
+_UI_EXTERN int uiControlEnabledToUser(uiControl *c);
 
 _UI_EXTERN void uiUserBugCannotSetParentOnToplevel(const char *type);
 
--- a/ui.h
+++ b/ui.h
@@ -478,14 +478,92 @@ _UI_EXTERN void uiButtonOnClicked(uiButton *b, void (*f)(uiButton *b, void *data
  */
 _UI_EXTERN uiButton *uiNewButton(const char *text);
 
+
+/**
+ * A boxlike container that holds a group of controls.
+ *
+ * The contained controls are arranged to be displayed either horizontally or
+ * vertically next to each other.
+ *
+ * @struct uiBox
+ * @extends uiControl
+ */
 typedef struct uiBox uiBox;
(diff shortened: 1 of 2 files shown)


ui.h
```diff
@@ -94,183 +106,1424 @@ struct uiControl {
...
+ * @param margined `TRUE` to set a window margin, `FALSE` otherwise.
+ * @memberof uiWindow
+ */
 _UI_EXTERN void uiWindowSetMargined(uiWindow *w, int margined);
+
+/**
+ * Returns whether or not the window is user resizeable.
+ *
+ * @param w uiWindow instance.
+ * @returns `TRUE` if window is resizable, `FALSE` otherwise. [Default: `TODO`]
```
mohad12211, line 410: I believe this commit https://github.com/libui-ng/libui-ng/commit/0713488ddf966727da844a52f24ea55e8620d660 sets resizeable on all platforms to be true

ui.h
```diff
@@ -94,24 +99,146 @@ struct uiControl {
...
+_UI_EXTERN void uiControlDisable(uiControl *c);
+
+/**
+ * Allocates a uiControl.
+ *
+ * Helper to allocate new controls.
+ *
+ * @param n Size of type to allocate.
+ * @todo Document parameters
+ * @memberof uiControl
```
msink, line 214: `@static @memberof`

All constructors have to be marked as static members

msink, line 214: `@static @memberof`

All constructors have to be marked as static members

szanni:
With regards to the callbacks: any ideas? I am reluctant to `typedef` these for the sole purpose of documentation.

My ideas are:

## 1
```c
/**
 * Registers a callback for ...
 *
 * @param x X instance.
 * @param f Callback function.
 * @param sender Back reference to the instance that initiated the callback.
 * @param senderData User data registered with the sender instance.
 * @param data User data to be passed to the callback.
 */
_UI_EXTERN void uiXOnChanged(uiX *x, void (*f)(uiX *sender, void *senderData), void *data);
```

Doxygen seems to be happy with that. Problem: how to document the return value as in:

```c
_UI_EXTERN void uiWindowOnClosing(uiWindow *w, int (*f)(uiWindow *w, void *data), void *data);      
```

Something like the following?
```c
/**
 * @param f Callback function.\n
 *          Returns:\n
 *          If @p f returns `TRUE`, the window is destroyed with the Destroy method.\n
 *          If @p f returns `FALSE`, or if OnClosing is never called, the window is not                      
 *          destroyed and is kept visible. 
 * @param sender Back reference to the instance that initiated the callback.
 * ...
 */
```
## 2

Document the entire callback function as part of the `@param f`, something like:
```c
/**
 * Registers a callback for ...
 *
 * @param x X instance.
 * @param f Callback function.\n
 *          @p sender Back reference to the instance that initiated the callback.\n
 *          @p senderData User data registered with the sender instance.\n
 *          Returns:\n
 *          If @p f returns `TRUE`, ...\n
 *          If @p f returns `FALSE`, ...
 * @param data User data to be passed to the callback.
 */
_UI_EXTERN void uiXOnChanged(uiX *x, int (*f)(uiX *sender, void *senderData), void *data);
```

cody271:
Option **2** definitely produces the best result, so lets go with that:

doxyclean_callbacks_option_2

mohad12211:
This is amazing work, are we missing something or is this ready to merge?

szanni:
> This is amazing work, are we missing something or is this ready to merge?

It is complete in of itself. It is still missing docs for quite a few API calls, especially around drawing and font rendering. The drawing API is something I would probably like to hand off to somebody else. I have never used a 2D library in my life, so my terminology is pretty much nonexistent. I'd rather have an expert take care of that one.

One thing I would still like to add is like a general description on initialization. A first steps essentially, something akin to what I did for uiTable. But this does not necessarily need to be in this PR.

I would mainly like somebody to look through the whole thing for errors. Yup, it is **massive**...

ui.h
```diff
@@ -285,6 +289,7 @@ _UI_EXTERN char *uiWindowTitle(uiWindow *w);
  * @param title Window title text.\n
  *              A valid, `NUL` terminated UTF-8 string.\n
  *              Data is owned by the caller.
+ * @note This method is merely a hint and may be ignored by the system.
```
bcampbell, line 292: Are there really any mainstream cases the title is ignored?
I know there must be window managers out there which don't show titles (eg WMs for kiosk use?).
But it seems to me that anyone affected would likely be purposefully running some setup which hides titles and not be surprised by such behaviour.
So this comment might be more confusing than helpful (to windows/mac developers, at least)?

bcampbell:
Fantastic work, by the way, @szanni - thanks so much for doing it!

ui.h
szanni: > Are there really any mainstream cases the title is ignored?

Not sure how mainstream my setup is, but I have been running i3wm (which is very popular) with the option `default_border pixel` for years. No title bars. I assume other titling WMs have similar options and for space saving purposes this would be reasonably common. Maybe that is not the case.

Thanks for looking through this btw!

szanni:
I removed the `[WIP]` in the title. The patchset is complete as it stands, I will not be adding more, unless something needs fixing.

More docs from my side will go into a new PR.

ui.h
bcampbell: Haha! Forgot to mention that I'm also an I3 user, although I've never fiddled much with the default config and I've never set mine up to hide window titles.
That was kind of my point: anyone who has configured their WM to not display titles probably isn't going to be too surprised that `uiWindowSetTitle()` doesn't have any visible effect, and the "this might be ignored" comment will just serve to alarm windows/mac/mainstream-WM-setup users.
Though... thinking about it, even if the title is hidden, the WM still probably stores it somewhere, right? So it's technically not ignored? Not 100% sure about that...

ui.h
```diff
@@ -94,24 +99,146 @@ struct uiControl {
...
+ * Returns the parent control.
+ *
+ * @param c uiControl instance.
+ * @returns The parent control, `NULL` if detached.
+ * @memberof uiControl
+ */
+_UI_EXTERN uiControl *uiControlParent(uiControl *c);
+
+/**
+ * Sets the control's parent.
```
bcampbell, line 139: Probably a good place to have a note on control ownership (a followup commit, maybe?)
Does calling `uiControlSetParent()` transfer ownership of the control to the parent? I _think_ it does, but I honestly can't remember!
Same goes for `uiControlDestroy()` - does it destroy child controls?

bcampbell (commented):
The text itself looks fine (I'm ignoring the doxygen `static`/`@memberof`/whatever details).
 
(@cody271 - don't want to step on your toes, but I figured you'd like to know there are other eyeballs on these commits, and doing a review seems like the way to indicate that. Let me know it you'd prefer something else)

ui.h
szanni: Very good points. I would like to do a writeup on ownership and such matters in a general section. Probably something like `@file` or so at the very top. Speaking of ownership, I am not sure if I am really happy with the current wording on string parameters.
I used the gtk docs as a template, but I don't find the wording to be that clear in the end. Does anyone have a nice set of phrases?
The problem I have with the phrasing right now is lifetimes. String parameters stay owned by the caller but are copied by the receiver, so the caller does not have to keep them alive. This is currently only implied and what users would expect, I guess? But then things get a little complicated with regards to `uiTableValue`, as there are some strange lifetime requirements.

Maybe something like:
`Data remains owned by the caller and is copied by the receiver.`
`Data is copied and ownership is transferred to the caller. Make sure to call uiFreeText()`

I always find these are big question marks in many projects and can not be spelled out explicitly enough. Do I need to free? Do I need to keep the reference alive (stack/heap)? etc.

> (edit: Oh. I was just looking at the uiControl commit... sorry, a bit new to the github review tools. I thought I could just review a single commit at a time)

Yes, this is very strange. I was sadly not aware of that, otherwise I would've tried to structure this differently. Because this is a huge patch set to review.

ui.h
cody271: Unfortunately there is no support in Doxygen for a special comment like `@constructor` for us to use.. 
`@static memberof` is the best answer here AFAICT.

We should probably add that to every `New` and `Alloc` API function. Definitely improves the output with that option:
uibutton

ui.h
```diff
@@ -94,222 +108,1827 @@ struct uiControl {
...
 _UI_EXTERN uiSlider *uiNewSlider(int min, int max);
 
+
+/**
+ * A control that visualizes the progress of a task via the fill level of a horizontal bar.
+ *
+ * Indeterminate values are supported via an animated bar.
+ *
+ * @struct uiProgressBar
+ * @extends uiControl
```
cody271, line 1201:
```suggestion
 * @extends uiControl
 * @ingroup static
```

ui.h
```diff
@@ -94,222 +108,1827 @@ struct uiControl {
...
+ * An exemplary, cross platform menu bar:
+ *
+ * - File
+ *   * New
+ *   * Open
+ *   * Save
+ *   * Quit, _use uiMenuAppendQuitItem()_
+ * - Edit
+ *   * Undo
+ *   * Redo
```
cody271, line 1772:
```suggestion
 *   * Redo
 *   * Cut
 *   * Copy
 *   * Paste
 *   * Select All
```

ui.h
```diff
@@ -1109,354 +2782,864 @@ struct uiAreaKeyEvent {
...
+ *
+ * uiTableValue objects are immutable.
+ *
+ * uiTable and uiTableModel methods take ownership of the uiTableValue objects
+ * when passed as parameter. Exception: uiNewTableValueImage().
+ *
+ * uiTable and uiTableModel methods retain ownership when returning uiTableValue
+ * objects. Exception: uiTableValueImage().
+ *
+ * @struct uiTableValue
```
cody271, line 3151:
```suggestion
 * @struct uiTableValue
 * @extends uiControl
```

ui.h
```diff
@@ -94,222 +108,1827 @@ struct uiControl {
...
-_UI_EXTERN void uiWindowOnClosing(uiWindow *w, int (*f)(uiWindow *w, void *data), void *data);
-_UI_EXTERN void uiWindowOnFocusChanged(uiWindow *w, void (*f)(uiWindow *, void *), void *data);
-_UI_EXTERN int uiWindowFocused(uiWindow *w);
 
+/**
+ * A control that visually represents a button to be clicked by the user to trigger an action.
+ *
+ * @struct uiButton
+ * @extends uiControl
+ * @ingroup button
```
cody271, line 486:
```suggestion
 * @ingroup button dataEntry
```

ui.h
```diff
@@ -1109,354 +2782,864 @@ struct uiAreaKeyEvent {
...
-//
-// You do not create uiTableValues directly; instead, you create a
-// uiTableValue of a given type using the specialized constructor
-// functions.
-//
-// uiTableValues are immutable and the uiTableModel and uiTable
-// take ownership of the uiTableValue object once returned, copying
-// its contents as necessary.
+/**
+ * @defgroup table Tables
```
cody271, line 3110:
```suggestion
```

ui.h
```diff
@@ -7,6 +7,15 @@
 // - const-correct everything
 // - normalize documentation between typedefs and structs
 
+/**
+ * @defgroup container Container widgets
+ * @defgroup dataEntry Data entry widgets
+ * @defgroup static Static widgets
+ * @defgroup button Buttons
+ * @defgroup dialogWindow Dialog windows
+ * @defgroup menu Menus
```
cody271, line 16:
```suggestion
 * @defgroup menu Menus
 * @defgroup table Tables
```

ui.h
cody271: As to avoid any confusion, the `@note` here should explicitly say on `On unix platforms`.

ui.h
```diff
@@ -1109,354 +2782,864 @@ struct uiAreaKeyEvent {
...
+ * Make sure the uiTableModel columns return the right type, as specified in
+ * the `uiTableAppend*Column()` parameters.
+ *
+ * The `*EditableModelColumn` parameters typically point to a uiTableModel
+ * column index, that specifies the property on a per row basis.\n
+ * They can however also be passed two special values defining the property
+ * for all rows: `uiTableModelColumnNeverEditable` and
+ * `uiTableModelColumnAlwaysEditable`.
+ *
+ * @struct uiTable
```
cody271, line 3525:
```suggestion
 * @struct uiTable
 * @extends uiControl
```

cody271:
> (@cody271 - don't want to step on your toes, but I figured you'd like to know there are other eyeballs on these commits, and doing a review seems like the way to indicate that. Let me know it you'd prefer something else)
> 
Just doing a review works yeah, and is much appreciated. :)

szanni:
> Unfortunately there is no support in Doxygen for a special comment like @constructor for us to use..
@static memberof is the best answer here AFAICT.

I can definitely add an `@static` keyword to all the constructors. It does help visual output albeit being incorrect.

There is a feature request from 2013 in the doxygen bugtracker [here](https://github.com/doxygen/doxygen/issues/5125). So I believe unless one of us implements this, we won't have a correct or complete way of specifying things.

I am still going to sprinkle some TODOs into some sections as suggested.

ui.h
cody271: Still missing plain `uiButton` in the data entry list, has font and color buttons.

cody271 (approved):
LGTM

Issue #8, opened 2022-01-09
Master issue for merging old PRs


cody271:

cody271:
@szanni Added you to the table above for all of your old PRs, now that you're here. 👍 

Hope that's okay.

szanni:
@cody271 Yes, thank you! I will be reviewing and resubmitting all my PRs over the coming days.

Should I be resubmitting issues from other people as well? Or is referencing the old issue sufficient, as some of the PRs should resolve those issues.

cody271:
@szanni Good question actually. I leave that up to your best judgement, to be decided on an issue by issue basis. While it sure would be nice to address each and every old issue one by one, the old PRs are the priority.

cody271:
Current plan is: before moving forward with anything new, every single one of these 82 old PRs needs to be addressed *ASAP*. I think the majority of them are probably ready to merge with trivial changes.

cody271:
So that we didn't have to wait for each of the original authors to come back, I've actually been using the GitHub API to generate this table above and mechanically move the old PRs over to the new repo.

rubyFeedback:
I think focusing just on merging old patches alone may already be super-helpful for people to want to aggregate. Once they see movement and momentum I am pretty confident that more and more people come and assist, if they see this fork being an opportunity.

cody271:
> I think focusing just on merging old patches alone may already be super-helpful for people to want to aggregate. Once they see movement and momentum I am pretty confident that more and more people come and assist, if they see this fork being an opportunity.

That is the hope yes. 🙏

DonMartio:
It is great to see so much movement. I was looking into libui#546 but i'm not shure yet if using g_ascii_strcasecmp is the best solution here. Furthermore i was starting to check for some of this warnings
`suggest braces around empty body in an ‘if’ statement`
last week.

cody271:
@DonMartio Yeah, `g_ascii_strncasecmp()` is fine for now.

Currently `uiprivStricmp()` is only used [here](https://github.com/libui-ng/libui-ng/blob/master/common/attribute.c#L238). Though take a look at the [Windows implementation](https://github.com/libui-ng/libui-ng/blob/master/windows/text.cpp#L109) for the best way: it converts to UTF-16 and does Unicode aware comparison.

szanni:
[...]

PR #100, opened 2022-06-02
Center new windows on MacOS

adamierymenko:
In lieu of a more comprehensive solution adding this one line causes new windows on macOS to appear in the center of the screen instead of in the lower left hand corner.

### DIFF
--- a/darwin/window.m
+++ b/darwin/window.m
@@ -202,6 +202,7 @@ static void uiWindowShow(uiControl *c)
 
 	[w->window makeKeyAndOrderFront:w->window];
 	[w->window center];
+	[[NSApplication sharedApplication] activateIgnoringOtherApps:YES];
 }
 
 static void uiWindowHide(uiControl *c)


cody271:
> In lieu of a more comprehensive solution adding this one line causes new windows on macOS to appear in the center of the screen instead of in the lower left hand corner.

So a more comprehensive solution was proposed already in this [old PR here](https://github.com/andlabs/libui/pull/271), but I think this new PR is actually a better solution. Keep it simple. Thoughts?

(Reminder that we are tracking all outstanding old PRs [here](https://github.com/libui-ng/libui-ng/issues/8))

adamierymenko:
My vote would be for this kind of simplicity. I think the scope and goal of this library should be to provide a fast, lightweight, cross-platform BASIC user interface, not to try to replicate the scope of something like Qt or .NET MAUI. If you need customizability and lots of sophisticated stuff go to one of those.

darwin/window.m
```diff
@@ -202,6 +202,7 @@ static void uiWindowShow(uiControl *c)
 
 	[w->window makeKeyAndOrderFront:w->window];
 	[w->window center];
+	[[NSApplication sharedApplication] activateIgnoringOtherApps:YES];
```
cody271, line 205: Is this line still needed once your app is packaged and running inside a proper `.app` bundle?

adamierymenko: Yes, it seems to be needed. Otherwise *sometimes* the app pops up under everything else. I have seen it in other apps.

It should *perhaps* be split off into a function like `uiFocusApplication()` with documentation to the effect that this is currently a no-op on platforms other than macOS. That way if someone needs to do this on Windows they could implement that function for Windows etc.

adamierymenko: BTW here is the project I'm using this with: https://github.com/zerotier/DesktopUI

It currently contains a forked and slightly modified libui-ng.

PR #101, opened 2022-06-03
Ignore ERROR_SUCCESS in Windows InitCommonControlsEx()

adamierymenko:
As per typical Microsoft behavior, `InitCommonControlsEx` appears to sometimes fail with an error of ERROR_SUCCESS. The library totally works fine in this case. I have not investigated really deeply but it seems this fixes init in this case. Platform is Windows 10 x64.

It's being used from a Rust app so it's possible that this is happening because Rust's `winapi` crate is calling another initializer or something, but I don't know. Just know that the library works fine if you ignore an error of "SUCCESS."

### DIFF
--- a/windows/init.cpp
+++ b/windows/init.cpp
@@ -105,8 +105,12 @@ const char *uiInit(uiInitOptions *o)
 	ZeroMemory(&icc, sizeof (INITCOMMONCONTROLSEX));
 	icc.dwSize = sizeof (INITCOMMONCONTROLSEX);
 	icc.dwICC = wantedICCClasses;
-	if (InitCommonControlsEx(&icc) == 0)
-		return ieLastErr("initializing Common Controls");
+        if (InitCommonControlsEx(&icc) == 0) {
+            if (GetLastError() != ERROR_SUCCESS) {
+                return ieLastErr("initializing Common Controls");
+            }
+        }
+
 
 	hr = CoInitialize(NULL);
 	if (hr != S_OK && hr != S_FALSE)


windows/init.cpp
```diff
@@ -105,8 +105,12 @@ const char *uiInit(uiInitOptions *o)
...
 	icc.dwSize = sizeof (INITCOMMONCONTROLSEX);
 	icc.dwICC = wantedICCClasses;
-	if (InitCommonControlsEx(&icc) == 0)
-		return ieLastErr("initializing Common Controls");
+        if (InitCommonControlsEx(&icc) == 0) {
+            if (GetLastError() != ERROR_SUCCESS) {
+                return ieLastErr("initializing Common Controls");
+            }
+        }
+
```
bcampbell, line 113: Looks good to me. I'd probably add a comment explaining it. Is there any info out there which documents this oddness? 
nit: indenting mismatch?

cody271: > Looks good to me. I'd probably add a comment explaining it. Is there any info out there which documents this oddness? nit: indenting mismatch?

Yup this needs:
1) comment with documentation / references to this issue.
2) formatting fixed to match existing style.

Otherwise looks good.

cody271 (changes requested):
This is also failing CI, due to the odd formatting issue?
```
../windows/init.cpp:108:5: error: '\U000000a0' was not declared in this scope
  108 |         if (InitCommonControlsEx(&icc) == 0) {
      |     ^
```

windows/init.cpp
adamierymenko: I couldn't find any docs in MSDN about this. I just found it via testing. On Windows 10 x64 init failed, and on close inspection init was failing but with a failure error code of "SUCCESS." I have however seen this weird quirk before in Windows APIs. Sometimes on Windows apps you will actually get error popups with a message like "Error: SUCCESS" for this (stupid) reason.

szanni:
> It's being used from a Rust app so it's possible that this is happening because Rust's `winapi` crate is calling another initializer or something, but I don't know. Just know that the library works fine if you ignore an error of "SUCCESS."

Maybe it is because of an upstream bug in rust? Is this reproducible with just a plain C version?

cody271:
> 
> Maybe it is because of an upstream bug in rust? Is this reproducible with just a plain C version?

This seems most likely.. Does Rust assume that all crate initialization is safely idempotent??
`uiInit()` definitely is not: it can only be called _once_ per process, from the main thread only.

adamierymenko:
Yeah, Rust probably assumes it's idempotent. In any case it seems the (undocumented?) Windows behavior is to return ERROR_SUCCESS on redundant init calls. Typical insane Windows API behavior.

adamierymenko:
It's probably good to be defensive here and ignore ERROR_SUCCESS since it will not be uncommon for this to be linked into apps that also may link things that carelessly assume these functions are idempotent.

cody271:
Instead of just Windows, see fix for all platforms: [here](https://github.com/libui-ng/libui-ng/pull/117).

cody271:
Duplicate of #117

szanni:
> Duplicate of #117

Not sure if this is the issue, or even an upstream rust one.

I just found this [PR in the old repo](https://github.com/andlabs/libui/pull/436) while going through older PRs. Looks _very_ similar to me. @adamierymenko are you adding a manifest to your binary as described in that PR?

PR #102, opened 2022-06-03
Position caret at end after uiEntrySetText() on Windows, replicating …

adamierymenko:
…behavior on MacOS

Just that. A few lines.

### DIFF
--- a/windows/entry.cpp
+++ b/windows/entry.cpp
@@ -66,6 +66,10 @@ void uiEntrySetText(uiEntry *e, const char *text)
 	// doing this raises an EN_CHANGED
 	e->inhibitChanged = TRUE;
 	uiWindowsSetWindowText(e->hwnd, text);
+        int l = (int)strlen(text);
+        if (l >= 0) {
+            SendMessage(e->hwnd, EM_SETSEL, l, l);
+        }
 	e->inhibitChanged = FALSE;
 	// don't queue the control for resize; entry sizes are independent of their contents
 }
--- a/windows/entry.cpp
+++ b/windows/entry.cpp
@@ -63,13 +63,12 @@ char *uiEntryText(uiEntry *e)
 
 void uiEntrySetText(uiEntry *e, const char *text)
 {
+	int l;
 	// doing this raises an EN_CHANGED
 	e->inhibitChanged = TRUE;
 	uiWindowsSetWindowText(e->hwnd, text);
-        int l = (int)strlen(text);
-        if (l >= 0) {
-            SendMessage(e->hwnd, EM_SETSEL, l, l);
-        }
+        l = (int)strlen(text);
+        SendMessage(e->hwnd, EM_SETSEL, l, l);
 	e->inhibitChanged = FALSE;
 	// don't queue the control for resize; entry sizes are independent of their contents
 }
--- a/windows/entry.cpp
+++ b/windows/entry.cpp
@@ -64,10 +64,12 @@ char *uiEntryText(uiEntry *e)
 void uiEntrySetText(uiEntry *e, const char *text)
 {
 	int l;
+	if (!text)
+		text = "";
 	// doing this raises an EN_CHANGED
 	e->inhibitChanged = TRUE;
 	uiWindowsSetWindowText(e->hwnd, text);
-        l = (int)strlen(text);
+	l = (int)strlen(text);
         SendMessage(e->hwnd, EM_SETSEL, l, l);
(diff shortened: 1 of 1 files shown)


cody271 (changes requested):
Same failure in CI due to odd formatting/encoding with this PR as well:
```
../windows/entry.cpp: In function 'void uiEntrySetText(uiEntry*, const char*)':
../windows/entry.cpp:69:5: error: '\U000000a0' was not declared in this scope
   69 |         int l = (int)strlen(text);
      |     ^
../windows/entry.cpp:70:2: error: expected ';' before '\U000000a0'
   70 |         if (l >= 0) {
      |  ^~
      |  ;
```

windows/entry.cpp
```diff
@@ -66,6 +66,10 @@ void uiEntrySetText(uiEntry *e, const char *text)
 	// doing this raises an EN_CHANGED
 	e->inhibitChanged = TRUE;
 	uiWindowsSetWindowText(e->hwnd, text);
+        int l = (int)strlen(text);
+        if (l >= 0) {
```
szanni, line 70: Unless I am missing something: how would this condition ever be `false`? `strlen` returns `size_t`, so it is by definition always `>=0`.
I would generally prefer matching types too `size_t len = strlen(text);` and then use that.

szanni (changes requested):

windows/entry.cpp
szanni: Oh and variable declaration at the top of the function. To stay consistent,

adamierymenko:
Fixed but didn't make it size_t because I think Windows expects int.

windows/entry.cpp
```diff
@@ -63,9 +63,12 @@ char *uiEntryText(uiEntry *e)
 
 void uiEntrySetText(uiEntry *e, const char *text)
 {
+	int l;
 	// doing this raises an EN_CHANGED
 	e->inhibitChanged = TRUE;
 	uiWindowsSetWindowText(e->hwnd, text);
+        l = (int)strlen(text);
```
szanni, line 70: Indentation is one tab, not spaces. You can see the mismatch in the Github diff.

szanni (changes requested):

szanni:
> Fixed but didn't make it size_t because I think Windows expects int.

Yeah, well windows actually expects a signed and unsigned type. Great. `UINT_PTR` and `LONG_PTR` according to `WPARAM` and `LPARAM` as defined [here](https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types).
I guess `int` is fine. It somehow does not seem to matter.

Otherwise the code does what it is supposed to. Unix behavior is matching too,

windows/entry.cpp
```diff
@@ -63,10 +63,15 @@ char *uiEntryText(uiEntry *e)
...
 void uiEntrySetText(uiEntry *e, const char *text)
 {
+	int l;
+	if (!text)
+		text = "";
 	// doing this raises an EN_CHANGED
 	e->inhibitChanged = TRUE;
-	uiWindowsSetWindowText(e->hwnd, text);
-	e->inhibitChanged = FALSE;
+        uiWindowsSetWindowText(e->hwnd, text);
```
szanni, line 71: I am not sure what editor you are using, but it seems incapable of getting the indentation right?

szanni (changes requested):

windows/entry.cpp
adamierymenko: I was using the GitHub online web editor to make small PRs. Will probably ditch that and actually check out a fork and do it in a proper editor. I guess the online editor is broken and for some reason inserts pseudo-whitespace unicode.

szanni: Oh wow. That sucks. Seems to work here (Firefox 101 on Linux).

szanni (changes requested):
Yup, good. The white space/indentation issues are still there though. L69-73.

cody271 (approved):
Confirmed to be passing CI.. hopefully no more GitHub online editor formatting weirdness. 😕 

I will just fix manually in master after merge. LGTM.

szanni:
Sorry for opening this issue again: @adamierymenko what was your exact use case here or the problematic behavior?

Is my understanding correct that this was about the caret placement when the `uiEntry` already has focus? To place the caret at the end when focused?

Because if the `uiEntry` does not have the focus the caret is always placed where I click. Correct?

Thanks

Related to #232 and #233

PR #103, opened 2022-06-03
Add "secret" names to uiMenuAppendItem for Mac cut/copy/paste

adamierymenko:
Add some "secret" names to uiMenuAppendItem to allow the Edit menu to be created on macOS to enable cut/copy/paste properly.

### DIFF
--- a/darwin/menu.m
+++ b/darwin/menu.m
@@ -259,9 +259,27 @@ static uiMenuItem *newItem(uiMenu *m, int type, const char *name)
 		[m->menu addItem:item->item];
 		break;
 	default:
-		item->item = [[NSMenuItem alloc] initWithTitle:uiprivToNSString(name) action:@selector(onClicked:) keyEquivalent:@""];
-		[item->item setTarget:uiprivAppDelegate().menuManager];
-		[m->menu addItem:item->item];
+		if (!strcmp(name, "@macPaste")) {
+			item->disabled = NO;
+			item->item = [[NSMenuItem alloc] initWithTitle:uiprivToNSString("Paste") action:@selector(paste:) keyEquivalent:@"v"];
+			[m->menu addItem:item->item];
+		} else if (!strcmp(name, "@macCut")) {
+			item->disabled = NO;
+			item->item = [[NSMenuItem alloc] initWithTitle:uiprivToNSString("Cut") action:@selector(cut:) keyEquivalent:@"x"];
+			[m->menu addItem:item->item];
+		} else if (!strcmp(name, "@macCopy")) {
+			item->disabled = NO;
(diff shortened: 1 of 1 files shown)


cody271:
> Add some "secret" names to uiMenuAppendItem to allow the Edit menu to be created on macOS to enable cut/copy/paste properly.

So this one is a bit tricky.. Thinking about how best to address this, few thoughts here:

`uiEntry` based controls on macOS currently don't respond to the standard keyboard shortcuts by default.
Keyboard shortcuts should probably work out of the box on all platforms, without requiring any specific menu items to be populated. That is the expected UX here.

`uiMenuItem` API for binding to "special" menu items doesn't support those used in text editing. Currently only supports: `About`, `Quit`, and `Preferences`.

And along with the `Cut`, `Copy`, `Paste`, and `Select All` proposed in this PR, I think we would also want `Undo` and `Redo`?

These would be needed on all platforms, not a macOS specific API addition.

Text editing special menu items are also different from the first three items we already support, in that they have a platform specific default callback associated with them each.

szanni:
> `uiEntry` based controls on macOS currently don't respond to the standard keyboard shortcuts by default. Keyboard shortcuts should probably work out of the box on all platforms, without requiring any specific menu items to be populated. That is the expected UX here.

Fully agree. Not just `uiEntry` though, but all the other input controls too. Otherwise something like subclassing as described [here](https://stackoverflow.com/questions/970707/cocoa-keyboard-shortcuts-in-dialog-without-an-edit-menu) might be viable.

> And along with the `Cut`, `Copy`, `Paste`, and `Select All` proposed in this PR, I think we would also want `Undo` and `Redo`?

I would probably keep the edit and (`Undo` and `Redo`) controls separate. Does `Undo` and `Redo` not apply to the entire application and hence would in most cases need application specific handling? Like a drawing app undoing a line draw?

> These would be needed on all platforms, not a macOS specific API addition.

Exactly. This is IMO the correct approach. unix and windows systems have these controls out of the box via the right click menu. But I am not sure if every user is tech savvy enough to use those. I can definitely think of a family member that might only use the window `Edit` menu. So from an accessibility point of view the mac menu enforcement might actually not be too dumb.

> Text editing special menu items are also different from the first three items we already support, in that they have a platform specific default callback associated with them each.

We might want to rethink the `uiMenu` API in general. It seems a little quirky, providing maybe a little too much freedom?
I mean I hate the _we know better_ approach, but maybe this is one of those rare cases where a stronger guiding hand is actually in the interest of the user, as well as the developer.

cody271:
> The problem is that without this cut/copy/paste simply do not work in Mac apps. There is no way to fix this via the default libUI API. It's not a problem on Windows or Linux.

Fix for cut/copy/paste on Mac was merged into [`master`](https://github.com/libui-ng/libui-ng/commit/fd2365595fb1f80fb0e3d393f3423d0b32f2234b). The issue of adding these edit actions as more "special" menu item types although remains an open question..

PR #104, opened 2022-06-07
Fix keyboard shortcuts in uiEntry based controls on macOS

cody271:
See #103


adamierymenko:
Ooh this is perhaps a better solution, since sometimes you don't want menus.

cody271:
Yes this needs to be fixed first, regardless of any potential future changes to the menu API..

cody271:
Please test this PR without your previous Mac specific patch if possible.

cody271:
Replaced by #107

PR #107, opened 2022-06-08
darwin: Fix keyboard handling for edit events.

szanni:
Fix handling of keyboard shortcuts in the absence of an Edit menu:
cut, copy, paste, select all, undo, redo.

This is the most simple work around I could come up with. It should work not only for uiEntry but the other controls too, hence it's own file.

Related to #103
Replaces #104

Edit: unix and windows work out of the box and do not need any special handling.

### DIFF
--- a/darwin/event.m
+++ b/darwin/event.m
@@ -1,0 +1,36 @@
+#import "uipriv_darwin.h"
+
+/**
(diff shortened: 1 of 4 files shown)


szanni:
Added some performance improvements.

cody271 (approved):
LGTM

PR #110, opened 2022-06-10
Fix more unix warnings

cody271:
Fixes all the remaining warnings on multiple UNIX platforms.

See #84

### DIFF
--- a/unix/table.c
+++ b/unix/table.c
@@ -22,25 +22,6 @@ struct uiTable {
 	void *headerOnClickedData;
 };
 
-// use the same size as GtkFileChooserWidget's treeview
-// TODO refresh when icon theme changes
-// TODO doesn't work when scaled?
-// TODO is this even necessary?
-static void setImageSize(GtkCellRenderer *r)
-{
-	gint size;
-	gint width, height;
-	gint xpad, ypad;
-
-	size = 16;		// fallback used by GtkFileChooserWidget
-	if (gtk_icon_size_lookup(GTK_ICON_SIZE_MENU, &width, &height) != FALSE)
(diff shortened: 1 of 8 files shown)


unix/table.c
```diff
@@ -22,25 +22,6 @@ struct uiTable {
 	void *headerOnClickedData;
 };
 
-// use the same size as GtkFileChooserWidget's treeview
-// TODO refresh when icon theme changes
-// TODO doesn't work when scaled?
-// TODO is this even necessary?
-static void setImageSize(GtkCellRenderer *r)
```
szanni, line 29: We might need this for padding? Maybe just comment it out. Code definitely works on my system.
Refresh for icon theme change seems kind of nonsense though?

unix/text.c
```diff
@@ -1,4 +1,5 @@
 // 9 april 2015
+#include <strings.h>
 #include "uipriv_unix.h"
 
 char *uiUnixStrdupText(const char *t)
```
szanni, line 5: Maybe add a comment in the header that this is broken for multibyte unicode on darwin and unix?

szanni (changes requested):
- common/attrlist.c Signed off
- test/images.c Signed off, proper fix in #86 
- unix/drawtext.c Signed off
- unix/menu.c Signed off
- unix/table.c See comments
- unix/text.c See comments
- unix/uipriv_unix.h See comments

szanni:
This would also supersede [old repo #546](https://github.com/andlabs/libui/pull/546). Tagging for #8

szanni (approved):
Signed off

PR #111, opened 2022-06-12
windows: Add C-style include guards to header files.

szanni:
Amazed that this has not become a problem before.

I am planning to rip out all the other nonsense in the headers later.


cody271 (approved):
LGTM

Issue #106, opened 2022-06-07
Multi-line labels result in REALLY wide windows on Windows.


adamierymenko:
If you put a multi-line string into a label, this results in a REALLY wide window on MS Windows. I think it's calculating the string length of the entire string without taking into account carriage returns and sizing the window as if it's one big long line. Correct behavior would be to tokenize the string by CR/LF and take the maximum width to find the width of the label.

The macOS code seems to do the right thing already. Don't know about Linux.

Current workaround is to use a multi-line read-only text entry but this sort of doesn't look right due to control styling.

cody271:
> Current workaround is to use a multi-line read-only text entry but this sort of doesn't look right due to control styling.

Screenshots always helpful: ideally of both the misbehaving label, and your read only entry workaround attempt if possible.

rubyFeedback:
I guess it may help either way to add a screenshot, perhaps a simple sample code where we can compare between all three platforms; and then see how much the fix proposed above would change that.

cody271:
Fixed in #112.

PR #112, opened 2022-06-18
Possibly fix multiline labels on windows

mohad12211:
unix and darwin appear to have no problems with multiline labels (would like confirm that).
this is an attempt to solve that on windows and address https://github.com/libui-ng/libui-ng/issues/106. 
the first commit from https://github.com/andlabs/libui/pull/465 to calculate the correct width and solve https://github.com/libui-ng/libui-ng/issues/106
<details>
<summary>Screenshot and code example</summary>

```c
#include <stdlib.h>
#include <string.h>
#include "../../ui.h"

int onClosing(uiWindow *w, void *data)
{
	uiQuit();
	return 1;
}

int main(int argc, char *argv[])
{
	uiInitOptions o;
	memset(&o, 0, sizeof (uiInitOptions));
	if (uiInit(&o) != NULL)
		abort();
	uiWindow *w = uiNewWindow("Hello", 100, 500, 1);
	uiBox *hbox = uiNewHorizontalBox();
	uiWindowSetChild(w, uiControl(hbox));
	uiButton *b = uiNewButton("TEST");
	uiLabel *l = uiNewLabel("1\n\n\n\n\n1\n\n\n\n\n1\n\n\n\n\n1\n\n\n\n\n1\n\n\n\n\n1\n\n\n\n\n1");
	uiBoxAppend(hbox, uiControl(b), 1);
	uiBoxAppend(hbox, uiControl(l), 1);
	uiWindowOnClosing(w, onClosing, NULL);
	uiControlShow(uiControl(w));
	uiMain();
	return 0;
}
```

before (smallest window width):
image
after: 
image

</details>

second commit is to calculate the correct height based on how many lines will be in the label.

<details>

<summary>Screenshot and code example</summary>

if we change the label in controlgallery to:
```c
uiControl(uiNewLabel("This is a label.\nRight now,\nlabels can\nonly span\none line.")),
```
before change:
image

after change:
image

</details>

This only addresses `uiLabel`, what about `uiForm`? does it make sense to have a form with multiline label?

### DIFF
--- a/windows/text.cpp
+++ b/windows/text.cpp
@@ -48,29 +48,26 @@ int uiWindowsWindowTextHeight(HWND hwnd)
 	int lineCount = 1;
 
 	text = windowTextAndLen(hwnd, &len);
-	for (start = text; start != text + len; start++) {
-		if (*start == '\n')
+	for (start = text; start != text + len; start++)
+		if (*start == L'\n')
 			lineCount++;
-	}
 
+	uiprivFree(text);
 	return lineCount * labelHeight;
 }
 
 int uiWindowsWindowTextWidth(HWND hwnd)
 {
 	LRESULT len;
-	WCHAR *text;
+	WCHAR *text, *start, *end;
 	HDC dc;
 	HFONT prevfont;
 	SIZE size;
 
-	size.cx = 0;
-	size.cy = 0;
-	//save the max width of multiline text
-	auto maxWidth = size.cx;
+	// save the max width of multiline text
+	int maxWidth = 0;
 
 	text = windowTextAndLen(hwnd, &len);
-	WCHAR *start = text, *end = text;
 	if (len == 0)		// no text; nothing to do
 		goto noTextOrError;
 
@@ -89,26 +86,19 @@ int uiWindowsWindowTextWidth(HWND hwnd)
 	}
 
 	// calculate width of each line
-	while(start != text + len) {
-		while(*start == '\n' && start != text + len) {
+	start = end = text;
+	while (start != text + len) {
+		while (*start == L'\n' && start != text + len)
 			start++;
-		}
-		if(start == text + len) {
+		if (start == text + len)
 			break;
-		}
 		end = start + 1;
-		while(*end != '\n' && end != text + len){
+		while (*end != L'\n' && end != text + len)
 			end++;
-		}
-		if (GetTextExtentPoint32W(dc, start, end - start, &size) == 0) {
+		if (GetTextExtentPoint32W(dc, start, end - start, &size) == 0)
 			logLastError(L"error getting text extent point");
-			// continue anyway, assuming size is 0
-			size.cx = 0;
-			size.cy = 0;
-		}
-		if(size.cx > maxWidth) {
+		else if (size.cx > maxWidth)
 			maxWidth = size.cx;
-		}
 		start = end;
 	}
 
--- a/windows/text.cpp
+++ b/windows/text.cpp
@@ -48,8 +48,11 @@ int uiWindowsWindowTextWidth(HWND hwnd)
 
 	size.cx = 0;
 	size.cy = 0;
+	//save the max width of multiline text
+	auto maxWidth = size.cx;
 
 	text = windowTextAndLen(hwnd, &len);
+	WCHAR *start = text, *end = text;
 	if (len == 0)		// no text; nothing to do
 		goto noTextOrError;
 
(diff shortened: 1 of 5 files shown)


windows/text.cpp
```diff
@@ -38,6 +38,24 @@ void uiFreeText(char *text)
...
+
+int uiWindowsWindowTextHeight(HWND hwnd)
+{
+	LRESULT len;
+	WCHAR* text, *start;
+	int lineCount = 1;
+
+	text = windowTextAndLen(hwnd, &len);
+	for (start = text; start != text + len; start++) {
+		if (*start == '\n')
```
cody271, line 52:
```suggestion
		if (*start == L'\n')
```

windows/text.cpp
```diff
@@ -66,20 +87,39 @@ int uiWindowsWindowTextWidth(HWND hwnd)
...
 	}
-	if (GetTextExtentPoint32W(dc, text, len, &size) == 0) {
-		logLastError(L"error getting text extent point");
-		// continue anyway, assuming size is 0
-		size.cx = 0;
-		size.cy = 0;
+
+	// calculate width of each line
+	while(start != text + len) {
+		while(*start == '\n' && start != text + len) {
```
cody271, line 93:
```suggestion
		while(*start == L'\n' && start != text + len) {
```

windows/text.cpp
```diff
@@ -66,20 +87,39 @@ int uiWindowsWindowTextWidth(HWND hwnd)
...
+			start++;
+		}
+		if(start == text + len) {
+			break;
+		}
+		end = start + 1;
+		while(*end != '\n' && end != text + len){
```
cody271, line 100:
```suggestion
		while(*end != L'\n' && end != text + len){
```

cody271 (commented):
This is comparing an ASCII literal against UTF-16 strings to parse the newlines. Otherwise looks good.

cody271:
> unix and darwin appear to have no problems with multiline labels (would like confirm that).

Confirmed to be a Windows only issue, correct.

> Screenshot and code example

This always makes reviewing much easier, thanks! 👍 

Maybe now also update that label text in the `controlgallery` example to reflect the fact that we _do_ support multiline labels on all platforms? ;)

> This only addresses `uiLabel`, what about `uiForm`? does it make sense to have a form with multiline label?

I wouldn't worry about supporting multiline labels in forms, no.

windows/text.cpp
```diff
@@ -48,8 +66,11 @@ int uiWindowsWindowTextWidth(HWND hwnd)
 
 	size.cx = 0;
 	size.cy = 0;
+	//save the max width of multiline text
+	auto maxWidth = size.cx;
 
 	text = windowTextAndLen(hwnd, &len);
+	WCHAR *start = text, *end = text;
```
szanni, line 73: Variable declaration always at the top.
And I would move the assignment just before the while loop, not somewhere in the middle of other code like so.
Assignment IMO looks nicer like this too: `start = end = text;`

windows/text.cpp
```diff
@@ -48,8 +66,11 @@ int uiWindowsWindowTextWidth(HWND hwnd)
 
 	size.cx = 0;
 	size.cy = 0;
+	//save the max width of multiline text
+	auto maxWidth = size.cx;
```
szanni, line 70:
```suggestion
	int maxWidth = 0;
```

windows/text.cpp
```diff
@@ -48,8 +66,11 @@ int uiWindowsWindowTextWidth(HWND hwnd)
 
 	size.cx = 0;
 	size.cy = 0;
```
szanni, line 68:
```suggestion
```
Not needed anymore.

windows/text.cpp
```diff
@@ -66,20 +87,39 @@ int uiWindowsWindowTextWidth(HWND hwnd)
...
+		}
+		if (GetTextExtentPoint32W(dc, start, end - start, &size) == 0) {
+			logLastError(L"error getting text extent point");
+			// continue anyway, assuming size is 0
+			size.cx = 0;
+			size.cy = 0;
+		}
+		if(size.cx > maxWidth) {
+			maxWidth = size.cx;
+		}
```
szanni, line 111:
```suggestion
		if (GetTextExtentPoint32W(dc, start, end - start, &size) == 0)
			logLastError(L"error getting text extent point");
		else
			if (size.cx > maxWidth)
				maxWidth = size.cx;
```
Looks clearer to me and removes a bunch of lines this way.

szanni (changes requested):
Nice work, definitely seems to work according to my tests.

Apart from removing out the old, now unneeded cruft, please remove the curly braces around the single if statements. The general guideline is to add them only when nesting gets too crazy or to match `if/else` blocks.

And please add the missing spaces around `if ()` and `while ()` statements. You did it in some places but not others.

Edit: and all the things @cody271 suggested of course.

windows/text.cpp
```diff
@@ -38,6 +38,24 @@ void uiFreeText(char *text)
...
+	LRESULT len;
+	WCHAR* text, *start;
+	int lineCount = 1;
+
+	text = windowTextAndLen(hwnd, &len);
+	for (start = text; start != text + len; start++) {
+		if (*start == '\n')
+			lineCount++;
+	}
+
```
szanni, line 55:
```suggestion
	uiprivFree(text);
```
Leaking memory here.

szanni (changes requested):
Missed the memory leak in my first review.

szanni (approved):
Very nice. Code looks good to me.

I tested things as well under Window 7 with assistive technologies, as I was a little concerned about the hard coded line height. Results:

100% font/icon size:
small

150% font/icon size:
big

It seems like Windows actually takes care of things via some type of smart scaling. So all good it seems.

PR #113, opened 2022-06-18
Fix meson builds with msvc to make them truly static.

### LINKED ISSUE #67: [Windows] Old libui may require MS Redistributable Package
Hi libui-ng developers! 

This is more of a report than an issue.

Ruby's libui binding packs the old libui shared library, and there were often reported cases where it did not work on Windows. Recently, one of the reasons for this was discovered: libui would not work unless `vcruntime140.dll` and `msvcp140.dll` was present. And to solve this problem, you should install the MS Redistributable Package on your Windows PC.

https://github.com/kojix2/LibUI/issues/48

In many cases, the MS Redistributable Package is installed as some kind of software bundle before the user can use libui. However, in new environments, the package may not be there.

To avoid this problem, Ruby's LibUI provides a Note in the README to install the MS Redistributable Package if it does not work on Windows. But If possible, I would like to run LibUI without installing MS Redistributable Package.

Thank you very much for taking over the maintenance of libui.

szanni:
Similar to mingw-w64 we need special flags to compile
the CRT in statically on static build with msvc.

Otherwise users need to install the redistributeable
package matching the compiler used for building,
making the build not full stand alone.

I hope this fixes #67. For me this patch definitely fixes the `vcruntime140.dll` error and I am able to run the resulting executables fully stand alone. I am unsure about `msvcp140.dll`  as I have never seen that one before. 

See:
https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -91,6 +91,20 @@ if libui_MSVC
 		'/MANIFEST:NO',
 	]
 
+	if libui_mode == 'shared'
+		if libui_is_debug
+			libui_project_compile_args += ['/MDd']
+		else
+			libui_project_compile_args += ['/MD']
+		endif
+	elif libui_mode == 'static'
+		if libui_is_debug
+			libui_project_compile_args += ['/MTd']
+		else
+			libui_project_compile_args += ['/MT']
+		endif
+	endif
+
 	# TODO autogenerate a .def file?
 else
 	libui_project_compile_args += [


meson.build
```diff
@@ -91,6 +91,20 @@ if libui_MSVC
...
+		else
+			libui_project_compile_args += ['/MD']
+		endif
+	elif libui_mode == 'static'
+		if libui_is_debug
+			libui_project_compile_args += ['/MTd']
+		else
+			libui_project_compile_args += ['/MT']
+		endif
+	endif
```
eli-schwartz, line 106: These options are all hooked up in Meson as `-Db_vscrt=`:
- none
- md
- mdd
- mt
- mtd
- from_buildtype
- static_from_buildtype

Specifically, what you're implementing here is identical to setting either `from_buildtype` (if libui_mode == 'shared') or `static_from_buildtype` (if libui_mode == 'static').

It's possible to set this in the project default_options, matching `'default_library=shared'`. People overriding that could just override both.

szanni: > Specifically, what you're implementing here is identical to setting either `from_buildtype` (if libui_mode == 'shared') or `static_from_buildtype` (if libui_mode == 'static').

Yeah, I did see that config option. Sadly I did not find a way to override the default options depending on build type. I would want it to automatically use `fom_buildtype` for `shared` and `static_from_buildtype` for `static`. I would expect `from_buildtype` to actually do that.

> It's possible to set this in the project default_options, matching `'default_library=shared'`. People overriding that could just override both.

This would require the user to pass another parameter `-Db_vscrt=`. I don't find this very user friendly. This is why I reimplemented the functionality. We are already doing the same for mingw builds anyways.

I am unsure why all those extra flags are needed in the first place. It is very unintuitive to me, why static ends up not being static. Not sure if these are ancient compiler/linker artifacts or if there is/was a good reason. For producing standalone binaries this is a nightmare though.

PR #114, opened 2022-06-18
Fix mingw warnings.

szanni:
Fix some of the mingw compile warnings. These are the straight forward ones.
There is still more to come, but I would like to look into those a little more.

The grid stuff can probably be just casted away, but I wonder why int and size_t is so mixed in the first place.

Related to #84

### DIFF
--- a/windows/attrstr.cpp
+++ b/windows/attrstr.cpp
@@ -335,7 +335,8 @@ static uiForEach processAttribute(const uiAttributedString *s, const uiAttribute
 		hr = p->layout->SetUnderline(hasUnderline, range);
 		if (hr != S_OK)
 			logHRESULT(L"error applying underline attribute", hr);
-		// and fall through to set the underline style through the drawing effect
+		// fall through to set the underline style through the drawing effect
+		// fall through
 	case uiAttributeTypeColor:
 	case uiAttributeTypeUnderlineColor:
 		// TODO const-correct this properly
--- a/windows/entry.cpp
+++ b/windows/entry.cpp
@@ -126,9 +126,11 @@ uiEntry *uiNewSearchEntry(void)
 	HRESULT hr;
 
 	e = finishNewEntry(0);
-	// TODO this is from ThemeExplorer; is it documented anywhere?
-	// TODO SearchBoxEditComposited has no border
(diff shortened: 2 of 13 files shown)


windows/entry.cpp
```diff
@@ -126,9 +126,11 @@ uiEntry *uiNewSearchEntry(void)
 	HRESULT hr;
 
 	e = finishNewEntry(0);
-	// TODO this is from ThemeExplorer; is it documented anywhere?
-	// TODO SearchBoxEditComposited has no border
-	hr = SetWindowTheme(e->hwnd, L"SearchBoxEdit", NULL);
-	// TODO will hr be S_OK if themes are disabled?
+
+	hr = SetWindowTheme(e->hwnd, L"SearchBoxEditComposited", NULL);
```
cody271, line 130: Why does this change the entry theme to `L"SearchBoxEditComposited"`?

windows/attrstr.cpp
```diff
@@ -335,7 +335,8 @@ static uiForEach processAttribute(const uiAttributedString *s, const uiAttribute
 		hr = p->layout->SetUnderline(hasUnderline, range);
 		if (hr != S_OK)
 			logHRESULT(L"error applying underline attribute", hr);
-		// and fall through to set the underline style through the drawing effect
+		// fall through to set the underline style through the drawing effect
+		// fall through
```
cody271, line 339: Having `fall through` comment here twice is intentional yes?

szanni: > Having `fall through` comment here twice is intentional yes?

Yep, sadly mingw only shuts up if the `// fall through` comment stands by itself :|

windows/entry.cpp
szanni: Ah yeah, guess that slipped in there. Well as the TODO suggested, use `L"SearchBoxEditComposited"` instead, as it does not exhibit a weird, non uniform indentation as `L"SearchBoxEdit"` does.

The current indentation does not line up (as seen in controllergallery) with the normal entry and password entry.

So essentially TODO removal, I could not see any regressions otherwise...

cody271 (approved):
LGTM

Issue #74, opened 2022-02-17
Automated test framework


szanni:
I would love to have some way of running automated tests, preferably some unit testing framework.
I am aware that a lot of things will still have to be tested manually, but at least covering some of the code base in an automated fashion would be an improvement.

I would suggest some type of unit testing framework over `assert()`. I know this most likely would mean introducing an external dependency. We could restrict usage to solely tests though.

Here a [list of unit testing framework on Wikipedia](https://en.wikipedia.org/wiki/List_of_unit_testing_frameworks#C). I would highly recommend a framework with mock support. From personal experience it becomes a pain really fast without.

I personally use cmocka, so my vote goes for that. In the end it doesn't really matter as long as the framework is mature.

rubyFeedback:
Guess you have to ask cody how he feels about unit testing (or other frameworks); and then other contributors. I don't know C well enough to be able to judge myself either way, but I think for any adoption there is some barrier or hurdle to overcome potentially. How may this impact new contributors who just want to quickly add some code/fix? Would they have to write new testing code first before able to send/submit patches or pull requests? Are these testing frameworks exclusive over one another? I guess the ideal would be to be able to retain the simplistic asserts() (if these are used; I did not check but I assume you did since you mentioned it in the report), and then "plug in" more test frameworks if need be or useful. Sort of like a plugin-like structure rather than a "one-size-fits-all" approach. These are just some semi-generic questions how it may impact or change a project. For instance, while testing is important, in ruby I often feel as if I lose too much time adding tests, in particular for simple code. For libraries used by many other people this may be different of course.

cody271:
> I would love to have some way of running automated tests, preferably some unit testing framework.

I'm very supportive of this idea. Testing is always good! The only reason it wasn't on my immediate roadmap is because of how to difficult it actually is in practice. Main issues are:

Even just adding basic unit testing with cmocka requires a lot of work, as libui-ng was written with the assumption of never running headless. Creating a virtual offscreen display within CI is absolutely doable, just very complex. Especially on all three platforms!

Unit testing only gets us so far. The real answer is to use a [graphical testing automation framework](https://en.wikipedia.org/wiki/Comparison_of_GUI_testing_tools). Again, testing on all three platforms with these tools is really quite challenging.

szanni:
[...]

PR #116, opened 2022-06-21
[POC] Unit testing with cmocka

szanni:
As I mentioned in #74 I would like some way of automatically testing at least some of the library code.

This is a first proof of concept, testing getters/setters, out of bounds values etc. I find this especially useful in making sure the API claims, eg. here #96 actually hold true.

~~You will need to install cmocka... via your package manager. Most repos seem to have it.
For mingw, install the appropriate mingw package. Otherwise this will fail horribly.~~

Current state:
- [x] unix
- [x] darwin
- [x] windows
  - [x] mingw
  - [x] msvc

# Update

Meson now force compiles `cmocka` locally, installed versions of cmocka should be ignored completely. This sadly requires bumping the meson version to 58.To test, run the resulting `unit` executable or `meson test -C build --verbose`.

~~macOS still fails after the first test due to #118.~~

## ~~unix~~

~~Install cmocka. Run the resulting `unit` executable.~~

## ~~darwin~~

~~Install cmocka. I did this via `brew`. Sadly brew does not install fat binaries, so I need to remove the `arm64` targets from the `meson.build`:~~

```patch
-       libui_project_compile_args += ['-arch', 'arm64', '-arch', 'x86_64']
-       libui_project_link_args += ['-arch', 'arm64', '-arch', 'x86_64']
+       libui_project_compile_args += ['-arch', 'x86_64']
+       libui_project_link_args += ['-arch', 'x86_64']
```

~~Again, run the resulting `unit` executable. It currently only runs the first test, then fails. I guess the uiUnInit() does not correctly do its job.~~
~~Can you open and close an cocoa app twice from the same process? I will try come up with some simpler test to verify and open an issue.~~

In theory we could probably reuse the same window over and over. But I think this might be a bug, so I'm leaving it in the POC.

### ~~mingw~~

~~Install mingw-w64-cmocka, or whatever the naming convention is on your system. Build. See it fail.

~~Currently meson decides to pull in the dynamic library of cmocka. The magic we need is in `test/unit/meson.build` where we actually want:~~
```patch
-cmocka_deps = dependency('cmocka', version : '>=1.0.0')
+cmocka_deps = dependency('cmocka', version : '>=1.0.0', static: true)
```
~~which sadly fails on my system. Cmocka seemingly does not produce a `libcmocka.a` but instead a `libcmocka-static.a`. I am unclear if this is a cmocka upstream issue or a problem with the Archlinux package I use.
Could somebody on another Linux distribution confirm?.~~

~~Or is there some way to coerce meson to go looking for a `lib-static.a` as well? Unclear. There seem to be quite a few open meson tickets regarding static linking.~~

~~Linking manually to `libcmocka-static.a` does work though, from the `build` directory:~~
```
x86_64-w64-mingw32-g++  -o meson-out/unit.exe meson-out/test_unit_resources.rc_resources.o meson-out/unit.exe.p/main.c.obj -flto -Wl,--allow-shlib-undefined -Wl,-O1 -static -static-libgcc -static-libstdc++ -Wl,-O1,--sort-common,--as-needed -fstack-protector -Wl,--start-group meson-out/libui.a -luser32 -lkernel32 -lgdi32 -lcomctl32 -luxtheme -lmsimg32 -lcomdlg32 -ld2d1 -ldwrite -lole32 -loleaut32 -loleacc -luuid -lwindowscodecs -L/usr/x86_64-w64-mingw32/lib -l:libcmocka-static.a -luser32 -lkernel32 -lgdi32 -lcomctl32 -luxtheme -lmsimg32 -lcomdlg32 -ld2d1 -ldwrite -lole32 -loleaut32 -loleacc -luuid -lwindowscodecs -Wl,--end-group -mconsole
```
~~and then running `./meson-out/unit.exe` does work with wine! Great.~~

### ~~msvc~~

~~I have no idea where to even start.~~

# ~~Conclusion~~

~~A lot to still be done. I am happy for any pointers. Looking at all the issues, I am wondering if pulling in `cmocka` via a git submodule and building it... might not be easier. As much as I had submodules, maybe it would keep the library more cohesive. especially with msvc in mind? Again: open for suggestions.~~


test/unit/meson.build
```diff
@@ -0,0 +1,24 @@
+cmocka_deps = dependency('cmocka', version : '>=1.0.0')
```
eli-schwartz, line 1:
```suggestion
cmocka_deps = dependency('cmocka', version : '>=1.0.0', default_options: ['default_library=static'])
```

And run `meson wrap install cmocka` to acquire the right file to make sure that Meson >= 0.55.0 can automatically fall back to downloading, checksumming, and building a private copy if it doesn't find a system one.

The default options ensures that when building as a subproject fallback it doesn't rather pointlessly build a shared library for cmocka.

eli-schwartz (changes requested):

test/unit/meson.build
szanni: Thanks for the suggestion, and great to see you here!

I tried that. It definitely works on darwin if I uninstall the cmocka brew package and remove the libui-ng fat binary hack.

Sadly is still fails with mingw.
If I leave the mingw-w64-cmocka package intact it will pull in the dynamic library.
If i uninstall it, it actually tries to pull in the linux headers in my case, but that might be a bug in `x86_64-w64-mingw32-meson` though.

eli-schwartz: > Thanks for the suggestion, and great to see you here!

I'm always happy to give people Meson tips. :)

> Sadly is still fails with mingw.

You can select to use the subproject even if mingw-w64-cmocka is installed. Meson provides fine-grained control for selecting between system and fallback versions:

```
# to force using fallbacks whenever possible, good for distributing standalone binaries that bundle dependencies
meson setup builddir/ --wrap-mode=forcefallback

# to force one specific fallback to be used
meson setup builddir/ --force-fallback-for=cmocka
```

I don't know what the linux headers issue is there. Logs might help I suppose -- because I don't have a Windows box, so I can't try running it to create those logs myself. :(

szanni: > ```
> # to force using fallbacks whenever possible, good for distributing standalone binaries that bundle dependencies
> meson setup builddir/ --wrap-mode=forcefallback
> 
> # to force one specific fallback to be used
> meson setup builddir/ --force-fallback-for=cmocka
> ```

Wow, yeah that actually works. Sweet. Still a lot of magic going on, but it looks like we can lock dependencies down and verify by hash.

> I don't know what the linux headers issue is there. Logs might help I suppose -- because I don't have a Windows box, so I can't try running it to create those logs myself. :(

This is actually from an ArchLinux box. mingw-w64 cross compile. Somehow the `mingw-w64-meson` package will use the systems `cmocka` library if it does not find a mingw  (`mingw-w64`) one. Which results in `-I/usr/include` erroneously getting passed to the compiler, causing havoc. Cmocka does not actually provide a proper way of building a static library and  `mingw-w64-cmocka` gets it wrong and produces a `cmocka-static.a`, which meson then can't find. I am not sure yet where to file a bug report yet. Probably in all projects.

eli-schwartz: > This is actually from an ArchLinux box. mingw-w64 cross compile. Somehow the `mingw-w64-meson` package will use the systems `cmocka` library if it does not find a mingw (`mingw-w64`) one. Which results in `-I/usr/include` erroneously getting passed to the compiler, causing havoc. Cmocka does not actually provide a proper way of building a static library and `mingw-w64-cmocka` gets it wrong and produces a `cmocka-static.a`, which meson then can't find. I am not sure yet where to file a bug report yet. Probably in all projects.

Oh yeah, that makes sense. :D I overlooked your above mention of Arch Linux and somehow thought you were using the MSYS2 shell. That package manager also has a mingw-w64-cmocka package...

So I *can*, as it happens, emulate your experience. However, what I discovered is as follows:
- if it is not installed, Meson does NOT pick up a system version via `/usr/bin/x86_64-w64-mingw32-pkg-config` (it does not detect my x86_64 linux version)
- if it *is* installed, it picks up the mingw sysroot version
  - but only the shared DLL. Unfortunately, the mingw sysroot isn't set up OOTB, so in this case you need to add `/usr/x86_64-w64-mingw32/bin` to $WINEPATH in order for cmocka.dll to be available when running the test program. If you do this, everything works.
  - the static lib is basically broken as you observed. I do not understand why upstream cmocka generates this broken name, it's totally wrong, but it's absolutely not the AUR package's fault, because upstream cmocka has an option -DBUILD_STATIC_LIB=ON that does: `add_library(cmocka-static STATIC ${cmocka_SRCS})`

msink: I created issue in MSYS2: https://github.com/msys2/MINGW-packages/issues/11912

msink:
On Windows box from MSYS2-MINGW64 console:
```
$ meson setup builddir --buildtype=release --default-library=static
...

$ ninja -C builddir --verbose
...
[98/115] "F:\msys64\mingw64\bin/windres.EXE" "-D_UI_STATIC" "../test/unit/resources.rc" "meson-out/test_unit_resources.rc_resources.o" "--preprocessor-arg=-MD" "--preprocessor-arg=-MQmeson-out/test_unit_resources.rc_resources.o" "--preprocessor-arg=-MFmeson-out/test_unit_resources.rc_resources.o.d"
[99/115] "cc" "-Imeson-out/unit.exe.p" "-Itest/unit" "-I../test/unit" "-Imeson-out" "-IF:/msys64/mingw64/include" "-fdiagnostics-color=always" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-Wextra" "-Wpedantic" "-std=c99" "-O3" "-Wno-unused-parameter" "-Wno-switch" "-D_UI_STATIC" -MD -MQ meson-out/unit.exe.p/main.c.obj -MF "meson-out/unit.exe.p/main.c.obj.d" -o meson-out/unit.exe.p/main.c.obj "-c" ../test/unit/main.c
...

$ builddir/meson-out/unit.exe
[==========] Running 11 test(s).
[ RUN      ] (sliderNew)
[       OK ] (sliderNew)
[ RUN      ] (sliderValueDefaultMin0)
[       OK ] (sliderValueDefaultMin0)
[ RUN      ] (sliderValueDefaultMin1)
[       OK ] (sliderValueDefaultMin1)
[ RUN      ] (sliderSetValue)
[       OK ] (sliderSetValue)
[ RUN      ] (sliderSetValueOutOfRangeClampLow)
[       OK ] (sliderSetValueOutOfRangeClampLow)
[ RUN      ] (sliderSetValueOutOfRangeClampHigh)
[       OK ] (sliderSetValueOutOfRangeClampHigh)
[ RUN      ] (sliderHasToolTipDefaultTrue)
[       OK ] (sliderHasToolTipDefaultTrue)
[ RUN      ] (sliderSetHasToolTip)
[       OK ] (sliderSetHasToolTip)
[ RUN      ] (sliderSetRangeLessThanValue)
[       OK ] (sliderSetRangeLessThanValue)
[ RUN      ] (sliderSetRangeGreaterThanValue)
[       OK ] (sliderSetRangeGreaterThanValue)
[ RUN      ] (sliderSetValueNoCallback)
[       OK ] (sliderSetValueNoCallback)
[==========] 11 test(s) run.
[  PASSED  ] 11 test(s).

```

but in native windows console it stiil needs `libcmocka.dll`
```
ntldd builddir/meson-out/unit.exe
        COMCTL32.dll => C:\Windows\WinSxS\amd64_microsoft.windows.common-controls_6595b64144ccf1df_5.82.7601.18837_none_a4d981ff711297b6\COMCTL32.dll (0x0000000000530000)
        d2d1.dll => C:\Windows\system32\d2d1.dll (0x0000000000eb0000)
        DWrite.dll => C:\Windows\system32\DWrite.dll (0x0000000000950000)
        GDI32.dll => C:\Windows\system32\GDI32.dll (0x0000000000eb0000)
        KERNEL32.dll => C:\Windows\system32\KERNEL32.dll (0x0000000000a30000)
        msvcrt.dll => C:\Windows\system32\msvcrt.dll (0x0000000000a30000)
        ole32.dll => C:\Windows\system32\ole32.dll (0x0000000001160000)
        USER32.dll => C:\Windows\system32\user32.dll (0x0000000000bb0000)
        libcmocka.dll => F:\msys64\mingw64\bin\libcmocka.dll (0x00000000000d0000)
```

eli-schwartz:
> I'm not expert in `meson`, adding `default_options: ['default_library=static']` to `cmocka_deps` did not change anything.

default_options controls the options used to build a subproject as a dependency fallback.

> ```
> Run-time dependency cmocka found: YES 1.1.5
> ```

This did not do a subproject, it used the default of what is on the system.

msink:
So - what should I change in `test/unit/meson.build` for statically linking `cmocka`?

szanni:
So I modified things similar to what @eli-schwartz suggested, but I decided to actually always compile `cmocka` locally, no matter what, via meson's `force_wrap` option.
Looking at the docs for meson's [subprojects](https://mesonbuild.com/Subprojects.html) I am probably still doing it wrong. But it works for now.

The decision to force build cmocka stems from all the problems arising due to broken or incomplete packaging.

- On Arch Linux the mingw-w64-cmocka library is not installed in the right place. Meson gets confused (I'll try to file a bug report upstream) and hence fails to fall back automatically.
- Meson has no concept of fat/universal binaries, so it will detect for example detect a `brew` package being present, but then fail when linking as the brew package only contains the native arch.
- No idea what Windows does with VS, and I am quite happy to not go explore that either.

I tested on Linux, wine (mingw), windows 7 VS2017, and macOS Big Sur.

All do compile and run. Mac still crashes after the first test due to #118, but it looks like things are moving forward.

And thanks @msink for testing ony msys2. I hope the new version works there too.

eli-schwartz:
> Looking at the docs for meson's [subprojects](https://mesonbuild.com/Subprojects.html) I am probably still doing it wrong. But it works for now.

This is not really so bad :) because you're just setting the project options to default to force the subproject. So if anyone actually wants to use the system version, it's easy to do so, they just need to pass a command line option to unset "force_fallback_for" -- there's no need to patch the source code or anything.

I wouldn't worry too much, I guess.

szanni:
> This is not really so bad :) because you're just setting the project options to default to force the subproject. So if anyone actually wants to use the system version, it's easy to do so, they just need to pass a command line option to unset "force_fallback_for" -- there's no need to patch the source code or anything.

That might actually be a feature, if people really want to use their local version for whatever reason.

szanni:
Another update:

- Unit tests can now be run through `meson test`.
- CI runs now run unit tests by default on all platforms. (It works! Exception being macOS due to #118)
- I added unit tests for `uiSlider` too and reorganized the code a little to give a structure for testing different uiControls.

I decided to have all unit tests for all uiControls in one binary. It looks like meson allows for as many binaries as we want - so in theory we could have a binary for each uiControl. cmocka is seemingly very good at failure recovery, so technically we do not need this. Any opinions on this?

All tests currently run in `<2s` on CI, on my system: `meson test -C build  0.49s user 0.11s system 112% cpu 0.533 total`.
I fear linking multiple binaries might be slower than the entire test run. Especially the mingw linker is excruciatingly slow.

eli-schwartz:
Meson doesn't care how many tests you have. But if you have all tests in one binary, they might all be reported as the same test PASS/FAIL status in `meson test`.

There's a generic mechanism for handling this sort of thing. https://testanything.org/

Meson supports it via `protocol: 'tap'` as a kwarg to `test()` and cmocka supports TAP as a reporting format if you set the `CMOCKA_MESSAGE_OUTPUT` environment variable. This isn't necessary, since a passing testsuite passes either way, and a failing testsuite fails either way, but it *would* mean that you get slightly nicer test harness reporting (specifically you can see the subtest count) in cases where some pass and some fail.

msink:
tested new version on my windows7 box, works perfectly.
`cmocka` was auto-dowlloaded, builded and linked statically:
```
MINGW32 /f/src/libui-ng
$ file builddir32/meson-out/unit.exe
builddir32/meson-out/unit.exe: PE32 executable (console) Intel 80386, for MS Windows

MINGW32 /f/src/libui-ng
$ ntldd builddir32/meson-out/unit.exe
        COMCTL32.DLL => C:\Windows\WinSxS\x86_microsoft.windows.common-controls_6595b64144ccf1df_5.82.7601.18837_none_ec86b8d6858ec0bc\COMCTL32.DLL (0x00500000)
        d2d1.dll => C:\Windows\system32\d2d1.dll (0x00bc0000)
        DWrite.dll => C:\Windows\system32\DWrite.dll (0x00bc0000)
        GDI32.dll => C:\Windows\system32\GDI32.dll (0x00490000)
        KERNEL32.dll => C:\Windows\system32\KERNEL32.dll (0x00c60000)
        msvcrt.dll => C:\Windows\system32\msvcrt.dll (0x00c60000)
        ole32.dll => C:\Windows\system32\ole32.dll (0x01010000)
        USER32.dll => C:\Windows\system32\user32.dll (0x00d10000)

MINGW64 /f/src/libui-ng
$ file builddir64/meson-out/unit.exe
builddir64/meson-out/unit.exe: PE32+ executable (console) x86-64, for MS Windows

$ ntldd builddir64/meson-out/unit.exe
        COMCTL32.dll => C:\Windows\WinSxS\amd64_microsoft.windows.common-controls_6595b64144ccf1df_5.82.7601.18837_none_a4d981ff711297b6\COMCTL32.dll (0x0000000000570000)
        d2d1.dll => C:\Windows\system32\d2d1.dll (0x0000000000f20000)
        DWrite.dll => C:\Windows\system32\DWrite.dll (0x0000000000a10000)
        GDI32.dll => C:\Windows\system32\GDI32.dll (0x0000000000610000)
        KERNEL32.dll => C:\Windows\system32\KERNEL32.dll (0x0000000000af0000)
        msvcrt.dll => C:\Windows\system32\msvcrt.dll (0x0000000000af0000)
        ole32.dll => C:\Windows\system32\ole32.dll (0x00000000011d0000)
        USER32.dll => C:\Windows\system32\user32.dll (0x0000000000c20000)
```

szanni:
macOS works now. For some reason a lot of the controls use some singleton delegate mechanism instead of subclassing (like other controls do), hence the singletons need to be destroyed on `uiUninit()`. I added a FIXME, but things do work nicely and it is a very straight forward, clean patch.

The other thing I changed it the way we include `cmocka`. I changed it around a bit by using `wrap_mode=forcefalback` by default. This should make it easier for package maintainers to switch to system dependencies for distributions that enforce offline builds and system packages. All in the README.

cody271 (approved):
LGTM

PR #117, opened 2022-06-24
Add idempotency guard to uiInit() on all platforms

cody271:


e1732a364fed:
please check this issue before merge

https://github.com/libui-ng/libui-ng/issues/160

PR #119, opened 2022-07-10
Export libui_dep for use of libui-ng as a meson subproject.

szanni:
Currently the meson.build does not export a variable so that the library can be used as a subproject.

The [meson docs](https://mesonbuild.com/Subprojects.html) suggest exporting a variable name `library_dep`. I went with `libui_dep` here as all the meson variables are called `libui_*`.

You can see an example project [here](https://github.com/szanni/sample-libui-ng). I'll update the URL to point the the main repo once this gets merged.

**Note:**

There are some other outstanding issues, namely that the parent project seemingly needs to define it's default options as `default_options: ['b_pch=false',  'c_std=c99', 'cpp_std=c++11']` (meson 0.62.2).

According to the [docs](https://mesonbuild.com/Subprojects.html) we should be able to pass `default_options` only to the subproject, but this does not work on my system for the options mentioned above. There are various closed bug reports like [this one](https://github.com/mesonbuild/meson/issues/2289).

With the newest meson (0.63.0) I can actually drop the C and C++ std options, but still need `b_pch`. I guesss that is an upstream bug. Is there a reason libui forces C headers to not be precompiled? Seems like a strange restriction? Is this just old cruft?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -180,6 +180,10 @@ if libui_mode == 'static'
 	libui_binary_deps = libui_deps
 endif
 
+# Export symbols for use of libui as a meson subproject
+libui_dep = declare_dependency(include_directories: include_directories('.'),
+	link_with : libui_libui)
+
 if get_option('tests')
 	subdir('test')
 endif


eli-schwartz:
> With the newest meson (0.63.0) I can actually drop the C and C++ std options, but still need `b_pch`. I guesss that is an upstream bug.

Yeah, as mentioned in the Meson ticket:
> There is no generic fix, for each option we want to make per-subproject we have to fix each place that option is accessed in Meson.

That being said, I wonder why this matters for pch? If libui doesn't define any `<lang>_pch: 'pch/foo_pch.h'` then libui won't use pch even if meson is configured to allow it.

szanni:
> That being said, I wonder why this matters for pch? If libui doesn't define any `<lang>_pch: 'pch/foo_pch.h'` then libui won't use pch even if meson is configured to allow it.

That's why I asked in the original comment. From my testing removing it makes no difference.

Other things are `warning_level`, `c_winlibs`, and `cpp_winlibs` . Why enforce them that way? If somebody wants to change these I see no reason why the library should flat out refuse to build.

The C and C++ versions make more sense. I sadly saw no way of specifying a _minimum_ C version in meson, or C features, because C11 should build just fine. I guess things might be a little trickier for C++ as they broke things in various places over the years. Not an expert on that though.

cody271 (approved):
LGTM

PR #120, opened 2022-07-11
windows: Remove unused winlibs.

szanni:
Remove the TODO to remove unused winlibs. These seem to be the minimum required libraries, or at least what I could come up with.


msink:
Since #125 `oleacc` is not unused too

PR #121, opened 2022-07-13
fix radiobutton unselected bug on unix

mohad12211:
calling `uiRadioButtonsSetSelected(-1)` in unix, instead of clearing the selection, won't do anything. That is because gtk doesn't allow a radiobutton group without a checked button, force unchecking a radiobutton won't work, so the workaround I suggested is to create a hidden button that is checked when calling `SetSelected(-1)`.

Also add bound checks to `uiRadioButtonsSetSelected()` for all platforms.

### DIFF
--- a/darwin/radiobuttons.m
+++ b/darwin/radiobuttons.m
@@ -172,9 +172,6 @@ void uiRadioButtonsSetSelected(uiRadioButtons *r, int n)
 	NSButton *b;
 	NSInteger state;
 
-	if (n < -1 || n > [r->buttons count] - 1)
-		uiprivUserBug("Index %d is out of bounds for uiRadioButtons with size %d.", n, [r->buttons count]);
-
 	state = NSOnState;
 	if (n == -1) {
 		n = uiRadioButtonsSelected(r);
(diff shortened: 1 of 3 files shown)


szanni:
Yup, works from my testing. A few things though:

Is there a reason to make the hidden button the last one? Why not the first? Does it mess with tab'ing through controls with the keyboard? Because making the hidden one first would elevate _all_ the special handling in getter and setter. For get just return `n-1` and for set, set `n+1`. No need to mess with the array bounds that way.

The other thing are the bounds checks. We don't have any right now. I would personally leave them out until we have a proper way or reporting errors.

And then there is the question of: do we want to support this if gtk purposefully suppresses this behavior? Probably yes from my side, but I believe it to be something to be discussed.

mohad12211:
> Is there a reason to make the hidden button the last one?

actually it was the first thing that I thought of for some reason, but yeah keeping it at the top is better.

> do we want to support this if gtk purposefully suppresses this behavior?

I can think of some use cases where you would want this, especially if you want no buttons selected initially.
I can also think of some reason to suppresses this behavior, but that's just an arbitrary rule IMO.

szanni (commented):
+12 -17. Fixed a bug and removed some code. Very nice!

I would still change the `hacky workaround` comment as indicated.

Otherwise signed off by me. Tab'ing works and the [unit tests](https://github.com/szanni/libui-ng/tree/cmocka-radio) now pass too.

szanni (approved):
Signed off. Nice work :)

PR #125, opened 2022-07-22
Add initial accessibility automation APIs

cody271:
Basic implementation for uiButton on unix and darwin, missing windows.

Includes new Page 18 in tester for a11y.

`uiA11y` is also added here as a "shadow" class: the type does not actually exist from the perspective of native C code, it is merely for organizational purposes in language bindings and documentation.

### DIFF
--- a/test/main.c
+++ b/test/main.c
@@ -52,6 +52,7 @@ int main(int argc, char *argv[])
 	uiBox *page15;
 	uiBox *page16;
 	uiBox *page17;
+	uiBox *page18;
 	uiTab *outerTab;
 	uiTab *innerTab;
 	int nomenus = 0;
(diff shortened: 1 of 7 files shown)


mohad12211:
I assume windows can be done like this?
```cpp
HWND btn;

btn = (HWND)uiControlHandle(uiControl(b));
SendMessageW(btn, BM_CLICK, 0, 0);
```
problem is [BM_CLICK](https://docs.microsoft.com/en-us/windows/win32/controls/bm-click) doesn't return anything to verify if the click was successful or not...

cody271:
Added windows implementation using [MSAA](https://en.wikipedia.org/wiki/Microsoft_Active_Accessibility) APIs. See the new "Page 18" in `tester` to try it out.

PR #133, opened 2022-08-11
test: Add manual testing suite for quality assurance.

szanni:
This is something that came up with the uiLabel multi line support for windows: a manual testing app.

I believe this to be useful for visual inspection as well as things we can not automate through a11y APIs.

I am aware the tester currently does that. The problem I have with the tester right now is: I have little to no idea what I should expect, unless I dig into the code of each component.

This here is an attempt to have a formal testing application for quality assurance (QA). This is something everybody can run to verify the workings of the library and have a more structured approach to reporting bugs. I think this would also be very helpful for identifying possible regressions in new PRs.

I am aware it will be a very complex task to complete the QA tester, but to be honest, it will be worth it. Working on the table selection API I already identified numerous upstream bugs through manual testing procedures (which I will make available soon).

Happy for feedback, especially with regards to the wording of the upstream bug reporting procedure on the `Quality Assurance` tab.

### DIFF
--- a/test/qa/main.c
+++ b/test/qa/main.c
@@ -1,0 +1,175 @@
+#include <stdio.h>
+
+#include "qa.h"
+
+struct controlTestCase {
+	const char *name;
+	uiControl* (*init)(void);
+	const char * (*guide)(void);
+};
+
+struct controlTestGroup {
+	const char *name;
+	struct controlTestCase *testCases;
+};
+
+#define QA_TEST(desciption, name) { desciption, name, name##Guide }
+
+struct controlTestCase labelTestCases[] = {
+	QA_TEST("1. Multi Line", labelMultiLine),
+	{NULL, NULL, NULL}
+};
+
+struct controlTestGroup controlTestGroups[] = {
+	{"uiLabel", labelTestCases},
+};
+
+uiControl* qaGuide()
+{
+	uiBox *box;
+	uiMultilineEntry *guide;
+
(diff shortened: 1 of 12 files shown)


szanni:
Rework depends on fixes in #136.

szanni:
Reworked things. I like it as it stands. Comments?

test/qa/main.c
```diff
@@ -0,0 +1,173 @@
+#include <stdio.h>
+
+#include "qa.h"
+
+struct controlTest {
```
cody271, line 5:
```suggestion
struct controlTestSection {
```

test/qa/main.c
```diff
@@ -0,0 +1,173 @@
...
+	const char *name;
+	uiControl* (*init)(void);
+	const char * (*guide)(void);
+};
+
+struct controlToTest {
```
cody271, line 11: Names `controlTest` and `controlToTest` are slightly confusing. How about `controlTestSection` ?

szanni: Yeah, one of the two hard problems in computer science. I was struggling with this one as well.

How about `controlToTest` and `groupToTest`?

cody271: `controlTestGroup` 👍

szanni:
Added manual testing for `uiButtonOnClicked`, as we can't cover this in the automated tests.
This is a good reference commit too for understanding how to add new tests to `qa`.

PR #134, opened 2022-08-11
test: Fix unit test for uiLabel to match description.

szanni:
The `labelNewEmptyString` does the same as the exact same thing as `labelNew`, whereas it was supposed to test empty string. This patch fixes that.


cody271 (approved):
LGTM

PR #135, opened 2022-08-12
test+doc: uiCheckbox unit tests and doc fixup.

szanni:
Unit tests for uiCheckbox and slightly updated docs: now documents defaults and notes that only one callback can be registered.


cody271 (approved):
LGTM

PR #136, opened 2022-08-13
Add unit tests for uiCombobox and fix unix and darwin bugs.

szanni:
Add unit tests for uiCombobox and fix the inconsistent triggering of `uiComboboxOnSelected` on unix platforms.

Again some may be considered upstream Gtk bugs (clearing an empty combobox with no selection triggers an `OnSelected` signal) but others are not (by the Gtk idea of when to trigger signals). I'm still planing to report these upstream but from my other reports (including patches) feedback has been nonexistent. As the fixes are the same anyways, here the fixes.

### DIFF
--- a/darwin/combobox.m
+++ b/darwin/combobox.m
@@ -79,16 +79,34 @@ static void uiComboboxDestroy(uiControl *cc)
 void uiComboboxAppend(uiCombobox *c, const char *text)
 {
 	[c->pbac addObject:uiprivToNSString(text)];
+	uiControlEnable(uiControl(c));
 }
 
 void uiComboboxInsertAt(uiCombobox *c, int n, const char *text)
 {
+	int selected = uiComboboxSelected(c);
+
 	[c->pbac insertObject:uiprivToNSString(text) atArrangedObjectIndex:n];
+	uiControlEnable(uiControl(c));
+
+	if (n <= selected)
+		uiComboboxSetSelected(c, selected+1);
+	else
+		uiComboboxSetSelected(c, selected);
 }
 
 void uiComboboxDelete(uiCombobox *c, int n)
 {
+	int selected = uiComboboxSelected(c);
+
 	[c->pbac removeObjectAtArrangedObjectIndex:n];
+
+	if (n < selected)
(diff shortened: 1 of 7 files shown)


szanni:
Looks like there are more issues on darwin. I will look into that a little later.

szanni:
Findings on darwin: apparently we want to support duplicate items, I personally find that very confusing, so fully agree with the macOS behavior of not allowing that, but ok:

```c
        // NSPopUpButton doesn't work like a combobox                                                   
        // - it automatically selects the first item                                                    
        // - it doesn't support duplicates                                                              
        // but we can use a NSArrayController and Cocoa bindings to bypass these restrictions
```
This means there is a custom workaround in place, that was apparently never properly tested. I hope the unit tests cover all possible scenarios and I have fixed all problems in the existing code.

I did one other modification for empty uiComboboxes: disable the control on darwin, matching the unix behavior. Otherwise I can visually select an empty field, which I personally find very confusing. Test on `page4`.

Regarding the workaround in general:

- According to the [docs](https://developer.apple.com/documentation/appkit/nspopupbutton/1534134-indexofselecteditem?language=objc) NSPopUpButton definitely supports no selection `-1`, so I am unsure if this would need to be set to `-1` by default, or if there are more issues.
- The no duplicates rule does indeed exist, I honestly can not think of a use case, where I would want the exact same item twice in a drop down: which one do I choose? Apple believes the same. But maybe there is a good reason to support that. I believe that if we would want to change that, we might want to do so in a different PR.

szanni:
Just documented the behavior of `uiComboboxDelete` for when the item deleted is the item selected, behavior being:
move the selection to the next item (if such an item exists) or `-1`.

I would personally probably always set it to `-1`.
Reasoning: the item has been deleted, so my selection should be invalidated.
Windows and unix do not do things in this way, but honestly win32 does not really do anything except for providing a skeleton to build upon and GTK is seriously broken in many regards, so not if these are good APIs to be imitated.

Apart from that, this PR should be complete.

cody271 (approved):
LGTM

Issue #141, opened 2022-09-29
How to get a version number


kojix2:
Hi!
I recently received a request to know the version of libui / libui-ng that you are using.
I replied that I will not support this in Ruby.

https://github.com/kojix2/LibUI/issues/63

However, being able to get a version number may be useful for binding authors and users.
In fact, when I create Ruby bindings for C libraries, my first goal is often to get the version.

I looked for a while, but the version number doesn't seem to be in the code. (Actually, it might be there.)
What do you think?

cody271:
> In fact, when I create Ruby bindings for C libraries, my first goal is often to get the version.
> 
> I looked for a while, but the version number doesn't seem to be in the code. (Actually, it might be there.)
> What do you think?
>
Correct, this is definitely an expected feature of libraries which is missing from `ui.h` API. Good catch!

szanni:
My suggested requirements on this would be:

- compile time
- string & integer

Good references IMO would be both [ffmpeg](https://ffmpeg.org/doxygen/trunk/group__lavu__ver.html) and [sqlite](https://www.sqlite.org/c3ref/c_source_id.html).

Something akin to:

- `UI_VERSION "x.y.z"`
- `UI_VERSION_MAJOR`, `UI_VERSION_MINOR`, `UI_VERSION_PATCH`
-  `UI_VERSION_NUMBER 00x00y00z` via `UI_VERSION_MAJOR * 1000 * 1000 + UI_VERSION_MINOR * 1000 + UI_VERSION_PATCH`

PR #142, opened 2022-09-29
Add uiVERSION

kojix2:
#141 

I have actually written code to solve this problem. I have very little experience with C, so I don't know if this is the right way to go about it, but it is better than not submitting any code at all.

It looks like the famous project is getting the version and generating version.h at build time. It is a great thing.

That is difficult for me, so I simply created version.c and wrote the version number there.

Thanks for maintaining libui-ng.

Since this is a prototype, you are, of course, free to close it. But if you improve it and merge it, that would be great.

### DIFF
--- a/common/meson.build
+++ b/common/meson.build
@@ -14,4 +14,5 @@ libui_sources += [
 	'common/tablevalue.c',
 	'common/userbugs.c',
 	'common/utf.c',
+	'common/version.c',
 ]
--- a/common/version.c
+++ b/common/version.c
@@ -1,0 +1,10 @@
+#include "../ui.h"
+#include "uipriv.h"
+
+#define UI_VERSION "ng-development-version"
+
+const char *uiVersion(void)
+{
+  static const char *libui_version_str = UI_VERSION;
+  return libui_version_str;
+}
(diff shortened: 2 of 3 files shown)


cody271:
> It looks like the famous project is getting the version and generating version.h at build time. It is a great thing.
> 
> Since this is a prototype, you are, of course, free to close it. But if you improve it and merge it, that would be great.
>
This needs to integrate with Meson to generate version.h at build time exactly, otherwise you have the right idea here.

szanni:
> > It looks like the famous project is getting the version and generating version.h at build time. It is a great thing.
> > Since this is a prototype, you are, of course, free to close it. But if you improve it and merge it, that would be great.
> 
> This needs to integrate with Meson to generate version.h at build time exactly, otherwise you have the right idea here.

I would actually say this is something that should be exposed in `ui.h` as a `#define` so it can be probed by developers at compile time. This is how all major libraries do things that I have worked with.

- `UI_VERSION "x.y.z"`
- `UI_VERSION_MAJOR`, `UI_VERSION_MINOR`, `UI_VERSION_PATCH`
-  `UI_VERSION_NUMBER 00x00y00z`

I am not fixed on the exact naming, but both [ffmpeg](https://ffmpeg.org/doxygen/trunk/group__lavu__ver.html) and [sqlite](https://www.sqlite.org/c3ref/c_source_id.html) for example look good to me.

kojix2:
Thank you all for your replies.
Since #143 is open, I will close #142.

PR #147, opened 2022-10-03
Impl. DPI awareness on Windows

norepimorphism:
This adds two new keys, `dpiAware` and `dpiAwareness`, to all of the Windows application manifests. `dpiAware` was introduced in Windows Vista and `dpiAwareness` is the recommended key on Windows 10. Microsoft docs suggest they should both be included for compatibility.

I've enabled per-monitor awareness, a description of which is available [here](https://learn.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows?redirectedfrom=MSDN#per-monitor-and-per-monitor-v2-dpi-awareness).

I also included the `supportedOS` key for Windows 10 to all of the application manifests. This is not strictly necessary for DPI awareness; I can remove this change and reintroduce it another PR if that is preferred.

I tested this by running `tester.exe` on my laptop&mdash;which has a Hi DPI monitor&mdash;before and after the change. I noticed that the application appeared blurry before the change but is clearer now.

Let me know if you run into any issues.

### DIFF
--- a/windows/draw.cpp
+++ b/windows/draw.cpp
@@ -26,18 +26,28 @@ ID2D1HwndRenderTarget *makeHWNDRenderTarget(HWND hwnd)
 {
 	D2D1_RENDER_TARGET_PROPERTIES props;
 	D2D1_HWND_RENDER_TARGET_PROPERTIES hprops;
+	HDC dc;
 	RECT r;
 	ID2D1HwndRenderTarget *rt;
 	HRESULT hr;
+
+	// we need a DC for the DPI
+	// we *could* just use the screen DPI but why when we have a window handle and its DC has a DPI
+	dc = GetDC(hwnd);
+	if (dc == NULL)
+		logLastError(L"error getting DC to find DPI");
 
 	ZeroMemory(&props, sizeof (D2D1_RENDER_TARGET_PROPERTIES));
 	props.type = D2D1_RENDER_TARGET_TYPE_DEFAULT;
 	props.pixelFormat.format = DXGI_FORMAT_UNKNOWN;
 	props.pixelFormat.alphaMode = D2D1_ALPHA_MODE_UNKNOWN;
-	props.dpiX = 0;
-	props.dpiY = 0;
+	props.dpiX = GetDeviceCaps(dc, LOGPIXELSX);
+	props.dpiY = GetDeviceCaps(dc, LOGPIXELSY);
 	props.usage = D2D1_RENDER_TARGET_USAGE_NONE;
 	props.minLevel = D2D1_FEATURE_LEVEL_DEFAULT;
+
+	if (ReleaseDC(hwnd, dc) == 0)
+		logLastError(L"error releasing DC for finding DPI");
 
 	uiWindowsEnsureGetClientRect(hwnd, &r);
 
@@ -67,9 +77,8 @@ ID2D1DCRenderTarget *makeHDCRenderTarget(HDC dc, RECT *r)
 	props.type = D2D1_RENDER_TARGET_TYPE_DEFAULT;
 	props.pixelFormat.format = DXGI_FORMAT_B8G8R8A8_UNORM;
 	props.pixelFormat.alphaMode = D2D1_ALPHA_MODE_PREMULTIPLIED;
-	// use default DPI
-	props.dpiX = 0;
-	props.dpiY = 0;
+	props.dpiX = GetDeviceCaps(dc, LOGPIXELSX);
+	props.dpiY = GetDeviceCaps(dc, LOGPIXELSY);
 	props.usage = D2D1_RENDER_TARGET_USAGE_GDI_COMPATIBLE;
 	props.minLevel = D2D1_FEATURE_LEVEL_DEFAULT;
 
--- a/windows/draw.cpp
+++ b/windows/draw.cpp
@@ -26,28 +26,18 @@ ID2D1HwndRenderTarget *makeHWNDRenderTarget(HWND hwnd)
 {
 	D2D1_RENDER_TARGET_PROPERTIES props;
 	D2D1_HWND_RENDER_TARGET_PROPERTIES hprops;
-	HDC dc;
 	RECT r;
 	ID2D1HwndRenderTarget *rt;
 	HRESULT hr;
-
-	// we need a DC for the DPI
-	// we *could* just use the screen DPI but why when we have a window handle and its DC has a DPI
-	dc = GetDC(hwnd);
-	if (dc == NULL)
-		logLastError(L"error getting DC to find DPI");
 
 	ZeroMemory(&props, sizeof (D2D1_RENDER_TARGET_PROPERTIES));
 	props.type = D2D1_RENDER_TARGET_TYPE_DEFAULT;
 	props.pixelFormat.format = DXGI_FORMAT_UNKNOWN;
 	props.pixelFormat.alphaMode = D2D1_ALPHA_MODE_UNKNOWN;
-	props.dpiX = GetDeviceCaps(dc, LOGPIXELSX);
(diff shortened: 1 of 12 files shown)


cody271 (commented):
This PR is dependent on the manifest file being correctly used and included in the build, which is not really guaranteed. Potentially would break in a  vendored copy of the lib inside an app, for example. I think this needs to attempt to enable HiDPI via the appropriate APIs at runtime as well.

Also missing some fixes for pixel calculations inside a few controls such as `uiColorButton`. See original HiDPI discussion: [here](https://github.com/andlabs/libui/issues/184).

norepimorphism:
I was not aware there was a previous HiDPI discussion; thanks for bringing it to my attention. It shouldn't be too difficult to additionally enable DPI awareness programmatically.

> Also missing some fixes for pixel calculations inside a few controls such as `uiColorButton`.

I'll take a look.

norepimorphism:
Yes, I can confirm there are differences in the sizes and positioning of some controls. For reference, I have compiled *libui-ng* with and without DPI awareness. See below for a visual comparison.

|Dpi Unaware|Dpi Aware|
|-----------|---------|
|A screenshot of the *libui-ng* Control Gallery "Basic Controls" tab without DPI awareness. It is blurry.|A screenshot of the *libui-ng* Control Gallergy "Basic Controls" tab with DPI awareness. The controls are rendered in crisp hi-definition, but the window is slightly smaller and the multiline entries are both narrower and shorter.|
|A screenshot of the *libui-ng* Control Gallery "Numbers and Lists" tab without DPI awareness. It is blurry.|A screenshot of the *libui-ng* Control Gallery "Numbers and Lists" tab with DPI awarenss. The image is clearer but the window is still smaller than the DPI-unaware version and the entries are slightly narrower.|
|A screenshot of the *libui-ng* Control Gallery "Data Choosers" tab without DPI awareness. It is blurry, but the controls are correctly positioned and, in particular, the color picker is correctly centered within its button control.|A screenshot of the *libui-ng* Control Gallery "Data Choosers" tab with DPI awareness. It is clearer but the window is smaller and the color picker is off-center, clipping through its containing button.|

The color picker is obviously wrong; it seems *libui-ng* thinks the size of the button is greater than it is, so it increases the left and top offsets such that the color preview rectangle now clips through the button.

Also, for some reason, the multiline entries are shorter in the DPI-aware build. I haven't looked at the code so I don't know if that's just an issue with how Control Gallery initializes the entries or if there's something awry with the multiline entry implementation itself. The other entry types look OK though.

norepimorphism:
I just pushed a commit that (I hope!) fixes the color button issue. Let me know if you notice any adverse effects, but I've run through a few of the example programs and it seems fine.

Here's a screenshot of the DPI aware build with the above fix:
image

windows/draw.cpp
```diff
@@ -77,8 +67,9 @@ ID2D1DCRenderTarget *makeHDCRenderTarget(HDC dc, RECT *r)
 	props.type = D2D1_RENDER_TARGET_TYPE_DEFAULT;
 	props.pixelFormat.format = DXGI_FORMAT_B8G8R8A8_UNORM;
 	props.pixelFormat.alphaMode = D2D1_ALPHA_MODE_PREMULTIPLIED;
-	props.dpiX = GetDeviceCaps(dc, LOGPIXELSX);
-	props.dpiY = GetDeviceCaps(dc, LOGPIXELSY);
+	// use default DPI
```
cody271, line 70: Won't this break mixed DPI setups?
Zero means the "system default DPI", which is the DPI of the first monitor detected in the system?

I had thought the purpose of querying the DC here was DPI awareness of the underlying monitor where the HWND is currently being displayed.

cody271:
See the original fixes for multiline entries and color buttons: [here](https://github.com/bennyhuo/libui/commit/741da3b41a459c731466a8419cc0e8f2cbea2eb2).

cody271:
> I've run through a few of the example programs and it seems fine.

What about `tester.exe`? I would expect some regressions on `Page 6`.

norepimorphism:
Sorry for disappearing for two months! I should be able to work on things for a little bit now.

> Won't this break mixed DPI setups? Zero means the "system default DPI", which is the DPI of the first monitor detected in the system?
> 
> I had thought the purpose of querying the DC here was DPI awareness of the underlying monitor where the HWND is currently being displayed.

I did not know that; thank you for letting me know. Though, I am curious why this seemed to work at all, or why the original code broke. I am not very familiar with DPI awareness or Win32 in general. From the original *libui* PR you linked to, it looks like Windows APIs assume a DPI of 92, or something? Is that related?

> What about `tester.exe`? I would expect some regressions on `Page 6`.

There are some minor differences, but I'm not sure what to look for, so I don't know if they're regressions or not. Below is a table of the differences between four of the pages. Let me know if you want me to provide more samples.

|Dpi Unaware|Dpi Aware|
|-----------|---------|
|Screenshot of "Page 6" with the "Original uiArea test" tab selected. It is blurry.|Screenshot of "Page 6" with the "Original uiArea test" tab selected.|
|Screenshot of "Page 6" with the "Arc test" tab selected. It is blurry.|Screenshot of "Page 6" with the "Arc test" tab selected.|
|Screenshot of "Page 6" with the "Direct2D: Creating a Simple..." tab selected. It is blurry.|Screenshot of "Page 6" with the "Direct2D: Creating a Simple..." tab selected.|
|Screenshot of "Page 6" with the "cairo samples: arc" tab selected. It is blurry.|Screenshot of "Page 6" with the "cairo samples: arc" tab selected.|

cody271:
> Though, I am curious why this seemed to work at all

It's sure a good question.. Whenever one attempts to revive an abandoned codebase, such questions inevitably will come up: `wait, how did this broken code ever work in the first place!?` 😛 

> There are some minor differences, but I'm not sure what to look for, so I don't know if they're regressions or not

Those screenshots all look correct to me so far. 👍 

This issue merits a lot more testing from a mixed DPI environment on Windows. That's probably the only way to resolve this.

norepimorphism:
OK, I implemented [the fix you linked](https://github.com/bennyhuo/libui/commit/741da3b41a459c731466a8419cc0e8f2cbea2eb2). The color button seems to still work, but I don't see any change in the multiline entry (even when I increase `WINVER` so that `GetDpiForWindow` is defined).

> This issue merits a lot more testing from a mixed DPI environment on Windows. That's probably the only way to resolve this.

I agree. I don't have a way to test that at the moment.

norepimorphism:
I do have a question, though. The fix uses `GetDpiForSystem` to scale the minimum size of multiline entries; I changed this to `GetDpiForWindow` because it sounds like `GetDpiForSystem` wouldn't work in a mixed-DPI system. But, wouldn't our approach in `draw.cpp` also work? (I.e., where we call `GetDeviceCaps(dc, LOGPIXELS{X,Y})`.) Or, do the values returned by `GetDpiForSystem` and `GetDeviceCaps` represent different things?

The advantage of possibly using `GetDeviceCaps` is that it isn't limited to Windows 10, which the `GetDpiForSystem` function (and friends) are.

cody271:
> I changed this to `GetDpiForWindow` because it sounds like `GetDpiForSystem` wouldn't work in a mixed-DPI system.

Yes `GetDpiForWindow` appears to be the correct solution in all cases.

> Or, do the values returned by `GetDpiForSystem` and `GetDeviceCaps(dc, LOGPIXELS{X,Y})` represent different things?

AFAICT from [the docs](https://learn.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows), these are functionally the same:
```
LOGPIXELSX | Number of pixels per logical inch along the screen width.
In a system with multiple display monitors, this value is the same for all monitors.
-- | --
LOGPIXELSY | Number of pixels per logical inch along the screen height.
In a system with multiple display monitors, this value is the same for all monitors.
```

PR #148, opened 2022-10-14
Add table on row clicked and double clicked callbacks.

szanni:
Proposal to add two new API functions to set a callback for when a user clicks and double clicks a table row.

```c
void uiTableOnRowClicked(uiTable *t, void (*f)(uiTable *t, int row, void *data), void *data);
void uiTableOnRowDoubleClicked(uiTable *t, void (*f)(uiTable *t, int row, void *data), void *data);
```

Implementations provided for darwin, unix, and windows. A small sample usage is provided on page16.

Two platform inconsistencies I found while testing:

1. Unix will send two `uiTableOnRowClicked` events when double clicking before the `uiTableOnRowDoubleClicked` event. Windows and Mac will only send one.
2. Mac will not select a row when ticking a checkbox, hence no `uiTableOnRowClicked` is sent. Unix and Windows select the row (first), send a `uiTableOnRowClicked` event and then send a uiTableValueChanged event.

My suggestions:

1. Can be ignored for now, as this will be automatically fixed once we switch to `GtkGesture` based events `Gtk >= 3.14` or at the latest Gtk4. A working fix that requires `Gtk >= 3.14` can be found [here](https://github.com/szanni/libui-ng/tree/table-on-row-clicked-and-double-clicked-gtk3-14).
2. Not sure what to do about the differing behavior for mac in general, this should possibly be raised as an issue on its own? Comments welcome.

This would supersede #44, adding a differentiation between single and double click and fixing all problems related to keyboard events on the unix side as well (by not implicitly adding any handlers).
Supersedes old PR [libui#516](https://github.com/andlabs/libui/pull/516)

### DIFF
--- a/unix/table.c
+++ b/unix/table.c
@@ -589,6 +589,27 @@ void uiTableOnRowDoubleClicked(uiTable *t, void (*f)(uiTable *, int, void *), vo
 	t->onRowDoubleClickedData = data;
 }
 
+#if GTK_CHECK_VERSION(3, 14, 0)
+static void onButtonPressed(GtkGestureMultiPress *gesture, gint nPress, gdouble wx, gdouble wy, gpointer data)
+{
+	uiTable *t = uiTable(data);
+	GtkTreePath *path;
+	gint row, x, y;
+
+	gtk_tree_view_convert_widget_to_bin_window_coords(t->tv, wx, wy, &x, &y);
+	gtk_tree_view_get_path_at_pos(GTK_TREE_VIEW(t->tv), x, y, &path, NULL, NULL, NULL);
+	if (path == NULL)
+		return;
+
+	row = gtk_tree_path_get_indices(path)[0];
+	gtk_tree_path_free(path);
+
+	if (nPress == 1)
+		(*(t->onRowClicked))(t, row, t->onRowClickedData);
+	else if (nPress == 2)
+		(*(t->onRowDoubleClicked))(t, row, t->onRowDoubleClickedData);
+}
+#else
 static gboolean onButtonPressed(GtkWidget *tv, GdkEventButton *event, gpointer data)
 {
 	uiTable *t = uiTable(data);
@@ -607,17 +628,21 @@ static gboolean onButtonPressed(GtkWidget *tv, GdkEventButton *event, gpointer d
 	row = gtk_tree_path_get_indices(path)[0];
 	gtk_tree_path_free(path);
 
-	if(event->type == GDK_BUTTON_PRESS)
+	if (event->type == GDK_BUTTON_PRESS)
 		(*(t->onRowClicked))(t, row, t->onRowClickedData);
-	else if(event->type == GDK_2BUTTON_PRESS)
+	else if (event->type == GDK_2BUTTON_PRESS)
 		(*(t->onRowDoubleClicked))(t, row, t->onRowDoubleClickedData);
 
 	return FALSE;
 }
+#endif
 
 uiTable *uiNewTable(uiTableParams *p)
 {
 	uiTable *t;
+#if GTK_CHECK_VERSION(3, 14, 0)
+	GtkGesture *gesture;
+#endif
 
 	uiUnixNewControl(uiTable, t);
 
@@ -636,7 +661,14 @@ uiTable *uiNewTable(uiTableParams *p)
 	// TODO set up t->tv
 	uiTableOnRowClicked(t, defaultOnRowClicked, NULL);
 	uiTableOnRowDoubleClicked(t, defaultOnRowDoubleClicked, NULL);
+
+#if GTK_CHECK_VERSION(3, 14, 0)
+	gesture = gtk_gesture_multi_press_new(GTK_WIDGET(t->tv));
+	g_signal_connect(gesture, "pressed", G_CALLBACK(onButtonPressed), t);
+	g_object_set_data_full(G_OBJECT(t->tv), "table-pressed-gesture", gesture, g_object_unref);
+#else
 	g_signal_connect(t->tv, "button-press-event", G_CALLBACK(onButtonPressed), t);
+#endif
 
 	gtk_container_add(t->scontainer, t->treeWidget);
 	// and make the tree view visible; only the scrolled window's visibility is controlled by libui
--- a/unix/table.c
+++ b/unix/table.c
@@ -20,6 +20,8 @@ struct uiTable {
 	guint indeterminateTimer;
 	void (*headerOnClicked)(uiTable *, int, void *);
 	void *headerOnClickedData;
+	void (*onRowDoubleClicked)(uiTable *, int, void *);
+	void *onRowDoubleClickedData;
 };
 
 /*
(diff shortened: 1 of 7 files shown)


cody271:
> My suggestions:
> 
> 1. Can be ignored for now, as this will be automatically fixed once we switch to `GtkGesture` based events `Gtk >= 3.14` or at the latest Gtk4. A working fix that requires `Gtk >= 3.14` can be found [here](https://github.com/szanni/libui-ng/tree/table-on-row-clicked-and-double-clicked-gtk3-14).

Our current Gtk dependency of `3.10` was released 2013, and `Gtk 3.14` was released 2014..
I'm fine with bumping our minimum Gtk version as part of this PR. Your `GtkGesture` based [fix](https://github.com/libui-ng/libui-ng/commit/3466eecdb75786e5164a969dcf86603857346592) looks correct to me.

```diff
diff --git a/unix/meson.build b/unix/meson.build
index f8004fbd..b1abd92a 100644
--- a/unix/meson.build
+++ b/unix/meson.build
@@ -49,7 +49,7 @@ libui_sources += [
 
 libui_deps += [
        dependency('gtk+-3.0',
-               version: '>=3.10.0',
+               version: '>=3.14.0',
                method: 'pkg-config',
                required: true),
```

cody271:
> 2. Not sure what to do about the differing behavior for mac in general, this should possibly be raised as an issue on its own? Comments welcome.
>
I think `uiprivOnCheckboxAction` on Darwin could be modified to emulate the Unix and Windows behavior?
Probably out of scope for this PR though yeah.

szanni:
> > My suggestions:
> > 
> > 1. Can be ignored for now, as this will be automatically fixed once we switch to `GtkGesture` based events `Gtk >= 3.14` or at the latest Gtk4. A working fix that requires `Gtk >= 3.14` can be found [here](https://github.com/szanni/libui-ng/tree/table-on-row-clicked-and-double-clicked-gtk3-14).
> 
> Our current Gtk dependency of `3.10` was released 2013, and `Gtk 3.14` was released 2014.. I'm fine with bumping our minimum Gtk version as part of this PR. Your `GtkGesture` based [fix](https://github.com/libui-ng/libui-ng/commit/3466eecdb75786e5164a969dcf86603857346592) looks correct to me.
> 
> ```diff
> diff --git a/unix/meson.build b/unix/meson.build
> index f8004fbd..b1abd92a 100644
> --- a/unix/meson.build
> +++ b/unix/meson.build
> @@ -49,7 +49,7 @@ libui_sources += [
>  
>  libui_deps += [
>         dependency('gtk+-3.0',
> -               version: '>=3.10.0',
> +               version: '>=3.14.0',
>                 method: 'pkg-config',
>                 required: true),
> ```

The other option would be to guard the fixed code in a `#if GTK_CHECK_VERSION(3, 14, 0)` and add a `@warning` in the API docs that two `uiTableOnRowClicked` events will be sent on unix, unless you manually want to bump the version number? (Haven't tested this yet). That way we would not need to bump the version number for the entire project. [Apparently](https://askubuntu.com/questions/638443/how-to-upgrade-gtk-3-10-to-gtk-3-14-on-ubuntu-14-04) Unity would break and first supported Ubuntu would be from 2015: 15.04.

Eventually we will want to bump this anyways, as a way of moving towards Gtk4. Question essentially is, do we want to put this on hold for the merge of API breaking old PRs or not.



> > 2. Not sure what to do about the differing behavior for mac in general, this should possibly be raised as an issue on its own? Comments welcome.
> 
> I think [`uiprivOnCheckboxAction`](https://github.com/libui-ng/libui-ng/blob/master/darwin/tablecolumn.m#L299) on Darwin could be modified to emulate the Unix and Windows behavior? Probably out of scope for this PR though yeah.

Yes, this will get especially tricky in conjunction with the multi select PR. I have not tested this yet, but something to look into is: how do systems modify an existing selection when ticking a checkbox with various keys held.

cody271:
> The other option would be to guard the fixed code in a `#if GTK_CHECK_VERSION(3, 14, 0)`
>
> Eventually we will want to bump this anyways, as a way of moving towards Gtk4. Question essentially is, do we want to put this on hold for the merge of API breaking old PRs or not.

Yeah that's a better option: delay the version bump a bit longer, if its not too much effort.

szanni:
Just added a fix that automatically gets applied when linking against `GTK >= 3.14`. All older unix systems still get the double trigger bug, which is documented as such in the API doc.

This sadly does introduce a new warning on newer systems, but should compile on older systems by using the `#else` branch (which is the important part). I believe the warning to be acceptable until we bump the version number.

```sh
../unix/table.c: In function ‘uiNewTable’:
../unix/table.c:666:9: warning: ‘gtk_gesture_multi_press_new’ is deprecated: Not available before 3.14 [-Wdeprecated-declarations]
  666 |         gesture = gtk_gesture_multi_press_new(GTK_WIDGET(t->tv));
      |         ^~~~~~~
In file included from /usr/include/gtk-3.0/gtk/gtk.h:117,
                 from ../unix/uipriv_unix.h:6,
                 from ../unix/table.c:2:
/usr/include/gtk-3.0/gtk/gtkgesturemultipress.h:45:14: note: declared here
   45 | GtkGesture * gtk_gesture_multi_press_new      (GtkWidget            *widget);
```

unix/table.c
```diff
@@ -636,7 +661,13 @@ uiTable *uiNewTable(uiTableParams *p)
 	// TODO set up t->tv
 	uiTableOnRowClicked(t, defaultOnRowClicked, NULL);
 	uiTableOnRowDoubleClicked(t, defaultOnRowDoubleClicked, NULL);
+
+#if GTK_CHECK_VERSION(3, 14, 0)
+	gesture = gtk_gesture_multi_press_new(GTK_WIDGET(t->tv));
```
cody271, line 666: Doesn't this leak the `GtkGesture` object? See [gtk_gesture_multi_press_new](https://docs.gtk.org/gtk3/ctor.GestureMultiPress.new.html) docs for ownership.

szanni: > Doesn't this leak the `GtkGesture` object? See [gtk_gesture_multi_press_new](https://docs.gtk.org/gtk3/ctor.GestureMultiPress.new.html) docs for ownership.

It does, my bad.
I somehow got confused by all the examples assigning the gesture object to some `priv` pointers that seem to get freed automagically.
Fixed it by following [this example](https://wiki.gnome.org/HowDoI/Gestures#Using_GtkGesture), as we do want to keep the default GtkTreeView handlers intact.

unix/table.c
```diff
@@ -563,9 +567,82 @@ static void uiTableDestroy(uiControl *c)
...
+
+	gtk_tree_view_convert_widget_to_bin_window_coords(t->tv, wx, wy, &x, &y);
+	gtk_tree_view_get_path_at_pos(GTK_TREE_VIEW(t->tv), x, y, &path, NULL, NULL, NULL);
+	if (path == NULL)
+		return;
+
+	row = gtk_tree_path_get_indices(path)[0];
+	gtk_tree_path_free(path);
+
+	if(nPress == 1)
```
cody271, line 607:
```suggestion
	if (nPress == 1)
```

unix/table.c
```diff
@@ -563,9 +567,82 @@ static void uiTableDestroy(uiControl *c)
...
+		(*(t->onRowClicked))(t, row, t->onRowClickedData);
+	if(nPress == 2)
```
cody271, line 609:
```suggestion
	else if (nPress == 2)
```

unix/table.c
```diff
@@ -563,9 +567,82 @@ static void uiTableDestroy(uiControl *c)
...
+		return FALSE;
+
+	gtk_tree_view_get_path_at_pos(GTK_TREE_VIEW(tv), event->x, event->y, &path, NULL, NULL, NULL);
+	if (path == NULL)
+		return FALSE;
+
+	row = gtk_tree_path_get_indices(path)[0];
+	gtk_tree_path_free(path);
+
+	if(event->type == GDK_BUTTON_PRESS)
```
cody271, line 631:
```suggestion
	if (event->type == GDK_BUTTON_PRESS)
```

unix/table.c
```diff
@@ -563,9 +567,82 @@ static void uiTableDestroy(uiControl *c)
...
+		(*(t->onRowClicked))(t, row, t->onRowClickedData);
+	else if(event->type == GDK_2BUTTON_PRESS)
```
cody271, line 633:
```suggestion
	else if (event->type == GDK_2BUTTON_PRESS)
```

cody271 (commented):
Nits

szanni:
Thanks for the nits, fixed :)

cody271 (approved):
LGTM

PR #149, opened 2022-10-16
ci: Add libtype shared to meson-windows

kojix2:
Hello. @cody271 @szanni and everyone.

I have a question and a pull request.

I would like to create a Ruby package containing a pre-release version of libui-ng.
But for now, I am attaching a Rake (Rake is the Ruby equivalent of Make) script that can download a build of libui-ng from the nightly.link you gave us so that people who want to use libui-ng can easily try it out.

However, on Windows, there is no shared library because shared is not enabled in the build.yml. Browsing through the blame, I noticed that @cody271 has intentionally disabled shared in the past.  (https://github.com/libui-ng/libui-ng/commit/9210b9f88ac524e3b5b2f55d8504827a13c3a2b9)

To see if it works, I ran ci with `shared` enabled in the forked repository of libui-ng. Then libui.dll was generated. 

https://github.com/kojix2/libui-ng/actions/runs/3258763203

The attached test file worked and I was able to call libui.dll from Ruby's LibUI. Of course I didn't look at all of them, and there may be something wrong with them, but at first glance they appear to be working.

So my question is: What is the reason that this option is disabled? 

Would you consider enabling this option if possible?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
         # windows-latest, windows-2022
         os: [windows-latest]
         arch: [x86, x64]
-        libtype: [static] # FIXME [static, shared]
+        libtype: [static, shared]
         # debug, debugoptimized, release
         buildtype: [debug]
     runs-on: ${{ matrix.os }}


szanni:
> The attached test file worked and I was able to call libui.dll from Ruby's LibUI. Of course I didn't look at all of them, and there may be something wrong with them, but at first glance they appear to be working.

Not sure what the reasoning was for disabling the shared build, as windows should work. Mingw is the build that is broken at the moment. @cody271 ?

The other thing you are probably aware of is the windows redistributable package that needs to be installed on the target machine (matching the compiler it was build with) as you noticed in #67. Did you ever get around to testing #113 and if that solves the problem?

cody271:
> So my question is: What is the reason that this option is disabled?

What appeared to be spurious failures in the Windows CI runners became too annoying is what I remember..
Though it's been a while so I'm not entirely sure, and we've improved the build system / CI config quite a bit since then.

Assuming we get another report of it working correctly now on Windows, I'm fine with re-enabling it in this PR.

kojix2:
Thank you for your response. I understand that it was previously disabled due to many pseudo failures in Windows CI.
However, as a binding producer, I would like to make it easy to try out the new libuii-ng. The more people who try libui-ng (knowing that it is a development version), the faster we will discover problems and new contributors.

@msink, @lolgab, @chopins, @rjopek, @zevv, @joaoventura, @MahmoudFayed, @norepimorphism, 
Thank you to all the active libui binding maintainers for watching this issue.

Is there anyone else who would like to get the latest libui-ng official shared library for Windows?
If so, could you please check if the Windows CI runner can generate the DLL?

PR #153, opened 2022-11-14
Fix meson subproject inclusion by making options project specific.

### LINKED ISSUE #152: darwin: Build fails when included as a meson subproject.
When including libui-ng as a meson subproject on macOS the build fails with:

```sh
subprojects/libui/meson.build:73:1: ERROR: Function 'add_global_arguments' cannot be used in subprojects because there is no way to make that reliable.
Please only call this if is_subproject() returns false. Alternatively, define a variable that
contains your language-specific arguments and add it to the appropriate *_args kwarg in each target.
```

You should be able to reproduce the issue by trying to build the [sample](https://github.com/szanni/sample-libui-ng) on macOS.

The offending section:
```meson
        libui_macosx_version_min = '-mmacosx-version-min=10.8' 
        add_global_arguments(libui_macosx_version_min, language: libui_darwin_langs)
        add_global_link_arguments(libui_macosx_version_min, language: libui_darwin_langs)

        libui_arch = ['-arch', 'x86_64', '-arch', 'arm64']
        add_global_arguments(libui_arch, language: libui_darwin_langs)
        add_global_link_arguments(libui_arch, language: libui_darwin_langs)
```

I would suggest switching at least to `add_project_arguments` but quite possibly these directives should be disabled completely when being built as a subproject? Especially the `arch` is something to decide for the superproject IMO.

szanni:
Use `add_project_*_arguments` instead of `add_global_*_arguments`.

Fixes #152

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -70,12 +70,12 @@ if libui_OS == 'darwin'
 	libui_darwin_langs = ['c', 'objc']
 
 	libui_macosx_version_min = '-mmacosx-version-min=10.8'
-	add_global_arguments(libui_macosx_version_min, language: libui_darwin_langs)
-	add_global_link_arguments(libui_macosx_version_min, language: libui_darwin_langs)
+	add_project_arguments(libui_macosx_version_min, language: libui_darwin_langs)
+	add_project_link_arguments(libui_macosx_version_min, language: libui_darwin_langs)
 
 	libui_arch = ['-arch', 'x86_64', '-arch', 'arm64']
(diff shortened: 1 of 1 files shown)


szanni:
The second commit adds a diff file to patch the cmocka `meson.build` and adds the minimum SDK version and fat binary directives manually. This would require meson 63.
It works, but I am unsure if this is the way to go. It makes it even more complicated for superprojects to change these setting, should they for example not want a fat binary.

Failing of the CI seems related to brew, not the patch from what I can tell.

PR #154, opened 2022-12-03
darwin: Remove uiButton singleton delegate by subclassing.

szanni:
Unclear to me what the whole delegate class story is all about.
Is that an artifact of the memory leak detector? IMO the current state of affairs is horrible.
I compiled and ran things with asan. All good.

### DIFF
--- a/darwin/button.m
+++ b/darwin/button.m
@@ -8,54 +8,39 @@ struct uiButton {
 	void *onClickedData;
 };
 
-@interface buttonDelegateClass : NSObject {
-	uiprivMap *buttons;
+@interface uiprivButton : NSButton {
+	uiButton *button;
 }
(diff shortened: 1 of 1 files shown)


szanni:
Spurious windows build fail.

szanni:
From a purist point of view this mixes view & controller. See my comment in #158

cody271:
> Spurious windows build fail.

I kicked it and CI seems to have passed now.

szanni:
> > Spurious windows build fail.
> 
> I kicked it and CI seems to have passed now.

Spurious build fails happen on my local machine, too, from time to time. Not sure if this is MSVC, ninja or meson.

szanni:
Found a way of triggering a crash without this patch, see my comment in #157

szanni:
Fixed in #176

Issue #157, opened 2022-12-04
darwin: Creation of some controls within a second window causes a crash.


szanni:
When creating certain controls within a second window through a user callback, the application crashes.

Here an example, or you can play around with the modified `hello-world` on this [branch](https://github.com/szanni/libui-ng/tree/darwin-uiNew-crash-on-user-callback).

```
#include <stdio.h>
#include <ui.h>

int onClosing(uiWindow *w, void *data)
{
	uiQuit();
	return 1;
}

int onClosingSub(uiWindow *w, void *data)
{
	return 1;
}

void buttonComboboxWindowOnClicked(uiButton *b, void *data)
{
	uiWindow *w;
	uiCombobox *child;
	w = uiNewWindow("Combobox", 300, 30, 0);
	uiWindowOnClosing(w, onClosingSub, NULL);
	child = uiNewCombobox();
	uiWindowSetChild(w, uiControl(child));
	uiControlShow(uiControl(w));
}

int main(void)
{
	uiInitOptions o = {0};
	const char *err;
	uiWindow *w;
	uiButton *b;

	err = uiInit(&o);
	if (err != NULL) {
		fprintf(stderr, "Error initializing libui-ng: %s\n", err);
		uiFreeInitError(err);
		return 1;
	}

	// Create a new window
	w = uiNewWindow("Hello World!", 300, 30, 0);
	uiWindowOnClosing(w, onClosing, NULL);

	b = uiNewButton("Combobox");
	uiButtonOnClicked(b, buttonComboboxWindowOnClicked, NULL);
	uiWindowSetChild(w, uiControl(b));

	/* Interestingly enough this works AND prevents the crash from happening
	 * when triggering the creating through the callback later on. */
	//buttonComboboxWindowOnClicked(uiButton *b, void *data)

	uiControlShow(uiControl(w));
	uiMain();
	uiUninit();
	return 0;
}
```
Crash log:

```
2022-12-04 00:29:41.337 hello-world[10896:501236] [libui] ../darwin/map.m:25:uiprivMapDestroy() POSSIBLE IMPLEMENTATION BUG; CONTACT ANDLABS:
attempt to destroy map with items inside
zsh: illegal hardware instruction  ./build/meson-out/hello-world
```

Controls I tested that exhibit that behavior:

- uiCombobox [showcase](https://github.com/szanni/libui-ng/tree/darwin-uiNew-crash-on-user-callback)
[...]

PR #158, opened 2022-12-05
darwin: Remove uiSlider singleton delegate.

szanni:
- Removes global state.
- Fixes `uiNewSlider()` crashing when attached to a secondary window that is being created via a user triggered callback.

See #157 

This removes the global uiSlider singleton delegate, which has been causing problems all over the place. It does so by subclassing NSSlider and assigning itself as the controller.
From a purist point of view this mixes view & controller. I personally don't mind, but I am happy to create a separate `uiprivSliderController` class and add an object to each uiSlider. This would of course incur a slight memory overhead. I guess that was what was trying to be avoided by the singleton in the first place?

### DIFF
--- a/darwin/slider.m
+++ b/darwin/slider.m
@@ -5,22 +5,6 @@
 // This will also be used for the initial frame size, to ensure the slider is always horizontal (see below).
 #define sliderWidth 92
 
-@interface libui_intrinsicWidthNSSlider : NSSlider
-@end
-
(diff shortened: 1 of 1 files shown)


cody271:
> From a purist point of view this mixes view & controller.

This is okay, since it is an internal implementation detail. Single subclass per control is cleanest.

szanni:
Fixed in #176

PR #159, opened 2022-12-05
darwin: Remove uiCombobox singleton delegate.

szanni:
- Removes global state.
- Fixes uiNewCombobox() crashing when attached to a secondary window that is being created via a user triggered callback. #157


szanni:
Fixed in #176

Issue #160, opened 2022-12-16
golang binding crash on darwin


e1732a364fed:
using libui-ng with andlabs/ui golang binding will crash with this message.

```
[libui] ../darwin/map.m:25:uiprivMapDestroy() POSSIBLE IMPLEMENTATION BUG; CONTACT ANDLABS:
attempt to destroy map with items inside
```

I found out it is related to this commit:

https://github.com/libui-ng/libui-ng/commit/f4d89db386ed882bec8a03d2c5e572f99aeaa800#diff-04bf4c4655bed893a14e225b932e29836d3d1f97c5e215e768e5270aa85f4d9e

it seems that the new approach in this commit will try to release object.
it's working on c programs.

But on the golang binding, the dealloc method is called somehow when it shouldn't, and it will result in the crash

 
I know very few about this. But I think I need to report this. Because it we don't fix this, no golang binding will work on darwin.

e1732a364fed:
By the way, currently the quick fix is just to run `git revert f4d89db`

cody271:
> Another question, is there any new golang binding projects?

See https://github.com/libui-ng/libui-ng/discussions/66

PR #162, opened 2022-12-17
darwin: Remove uiWindow singleton delegate by subclassing.

szanni:
- Removes global state and related problems

Related to #157 and #160

I hope I did not break any of the callback code or menu shenanigans. I tested some of it in the tester, but since we don't have an established protocol yet, it's hard to tell.

@e1732a364fed does that fix window creation in the golang bindings for you with this patch? Not sure if you can test.

### DIFF
--- a/darwin/uipriv_darwin.h
+++ b/darwin/uipriv_darwin.h
@@ -77,7 +77,16 @@ extern void uiprivFinishNewTextField(NSTextField *, BOOL);
 extern NSTextField *uiprivNewEditableTextField(void);
 
 // window.m
-@interface uiprivNSWindow : NSWindow
+@interface uiprivNSWindow : NSWindow<NSWindowDelegate> {
+	uiWindow *window;
+}
+- (BOOL)windowShouldClose:(id)sender;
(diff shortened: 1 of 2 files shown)


szanni:
Not sure how that typo slipped in. Should work now.

e1732a364fed:
I tested your PR. Not working on darwin. Same crash issue.

szanni:
> The uiWindow now works fine. But it is the other controls init that still make the program crash. So the #160 issue is still not fixed, because unless you want your user to only create an empty window with nothing inside.
> 
> for example, create an checkbox will still lead to crash on darwin in the golang binding.

Yes, that is sadly still to be expected. This needs to be fixed for all controls that use the global delegate mechanism.

I am glad that reverting the offending commit works for you, sadly it does break the testing infrastructure as well. I believe fixing the underlying issue to be the better course of action. I've already committed PRs for about half of the offending uiControls.

e1732a364fed:
> I've already committed PRs for about half of the offending uiControls.

Thanks for your hardwork!

PR #170, opened 2022-12-22
unit: Add unit tests for uiMenu

### LINKED ISSUE #169: uiNewMenu() fails when called a second time after uiUninit() and uiInit()
Similar to #118, another `uiUninit()` followed by `uiInit()` problem: Any application creating a menu and hence possibly calling `uiNewMenu()` after a reset (`uiUninit(); uiInit();`) will refuse to work the second time around.

You can check it out by running `meson test` on [this branch](https://github.com/szanni/libui-ng/tree/cmocka-menu)

Does anyone know why the menu needs to be finalized in the first place? Why can we not modify the menu after creating the first window? Is this related to a limitation of one of the underlying platforms or just a library quirk?

szanni:
Edit: I must have pressed the wrong PR button. I would still like to add more tests to this. But here you go.

Add unit tests for uiMenu. Depends on #171 

Edit2: [Branch](https://github.com/szanni/libui-ng/tree/fix-uiMenu-uninit-cmocka) That combines the fixes in #171 with the tests here.

### DIFF
--- a/test/unit/main.c
+++ b/test/unit/main.c
@@ -2,22 +2,11 @@
 
 #include "unit.h"
 
-#define UNIT_TEST_WIDTH 300
-#define UNIT_TEST_HEIGHT 200
-
-static int onClosing(uiWindow *w, void *data)
+int unitWindowOnClosingQuit(uiWindow *w, void *data)
 {
 	uiQuit();
 	return 1;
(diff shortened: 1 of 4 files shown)


szanni:
Fixed in #176

PR #171, opened 2022-12-22
Fix uiNewMenu when initializing libui-ng a second time.

### LINKED ISSUE #169: uiNewMenu() fails when called a second time after uiUninit() and uiInit()

szanni:
Reset the global variables on `uiprivUninit()` for all platforms.
This will should allow for re-initializing an application that contains menus.

From preliminary testing the code seems to work on all platforms. I tested Window7, Linux (gtk 3.24.35), macOS Big Sur. Hope I did not miss something.

Fixes #169

Edit: [Branch](https://github.com/szanni/libui-ng/tree/fix-uiMenu-uninit-cmocka) That combines the fixes here with the tests in #170.

### DIFF
--- a/darwin/menu.m
+++ b/darwin/menu.m
@@ -365,4 +365,8 @@ void uiprivUninitMenus(void)
 		uiprivFree(m);
 	}];
 	[menus release];
+
+	/* Reset global state. */
+	menus = nil;
+	menusFinalized = NO;
 }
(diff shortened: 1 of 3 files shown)


szanni:
Fixed in #176

Issue #165, opened 2022-12-17
darwin: Remove singleton delegates and associated global state


szanni:
This has been causing a lot of problems, both in C code #157 and #133, and in bindings #160.

IMO this needs to be resolved as soon as possible.

| control | possible fix | done |
|-|-|-|
| uiButton |#154 | |
| uiCheckbox | #166| |
| uiCombobox |#159 | |
|uiDatetimepicker | #168| |
|uiEditableCombobox|#167 | |
|uiEntry|#175 | |
|uiMenu |#172 | |
|uiSlider |#158 | |
|uiWindow |#162 | |

szanni:
Fix suggestions are up for all affected controls.

Frityet:
What problems are present as a result of the singletons?

PR #172, opened 2022-12-30
uiMenu: Fix uiUninit() + cmocka tests + darwin: Remove uiprivMap()

szanni:
- Fixes presented in #169 for all platforms pertaining uiMenu re-initialization
- Cmocka test cases for uiMenu #170
- Darwin: refactoring removing global state and uiprivMap() #165
- Darwin: fix memory leak of strings `"About unit"`, `"Quit unit"`

The existing memory leak around the NSObjects for `About`, `Preferences` and `Quit` still exists however.
Run `leaks --atExit -- ./build/meson-out/unit` to see what objects remain. Interestingly enough test case triggering the problem is `initUninitTwice`. I haven't been able to figure out who still holds a reference, I assume it has to do with the mixing of ARC, MRR and `@autoreleasepool`. But maybe the problem lies somewhere else, as the problem seemingly disappears when creating a window.

### DIFF
--- a/darwin/menu.m
+++ b/darwin/menu.m
@@ -365,4 +365,8 @@ void uiprivUninitMenus(void)
 		uiprivFree(m);
 	}];
 	[menus release];
+
+	/* Reset global state. */
+	menus = nil;
+	menusFinalized = NO;
 }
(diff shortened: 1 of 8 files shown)


szanni:
This seems to be broken. Last commit

szanni:
Fixed in #176

PR #179, opened 2023-03-24
Remove `SimplexiDev.UI` from the bindings list.

simplexidev:
I will be (eventually) creating a new project for C# bindings, but for now, the `sdfx` project has a different roadmap at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Language | Bindings
 --- | ---
 C++ | [libui-cpp](https://github.com/billyquith/libui-cpp), [cpp-libui-qtlike](https://github.com/aoloe/cpp-libui-qtlike)
 C# / .NET Framework | [LibUI.Binding](https://github.com/NattyNarwhal/LibUI.Binding)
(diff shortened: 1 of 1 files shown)


cody271:
So the language bindings section of the README intentionally includes many others that are also unmaintained or abandoned..that way we can keep track of everything in one place easily for now. I suggest that you mark your repo as [archived](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories) to indicate its current status instead. You can always [unarchive](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories#unarchiving-a-repository) at a later date.

simplexidev:
The repository itself is moving in a different direction, i.e. `libui` functionallity will be removed and replaced with pure C# for UI and controls (for the most part) wrapped over OpenGL for graphics and native windowing. The repository `sdfx` was renamed yesterday to `sdevlibs`, and will be removing `libui` API in future commits. The link will be invalid.

Issue #161, opened 2022-12-16
fix windows bad performace?


e1732a364fed:
related issues:

https://github.com/andlabs/libui/issues/311

https://github.com/andlabs/ui/issues/184

Since the old project is dead, I think we should pay attention to this too.

e1732a364fed:
here's my script for temporarily fix this 

```
import sys

def comment_out_lines(filename, string_to_search, end_string, comment_char):
  # Open the file in read mode
  with open(filename, 'r') as file:
    # Read all lines in the file
    lines = file.readlines()

  # Initialize the commenting flag
  commenting = False

  # Open the file in write mode
  with open(filename, 'w') as file:
    # Iterate over all lines
    for line in lines:
      # Check if the string is present in the line
      if string_to_search in line:
        # Set the flag to indicate that we are in the commented out section
        commenting = True
      
      # If we are in the commented out section, add the comment character at the beginning of the line
      if commenting:
        file.write(comment_char + line)
      # Otherwise, write the original line
      else:
        file.write(line)

      if end_string in line:
        # If the end string is found, reset the flag
        commenting = False


# Test the function
comment_out_lines('windows/parent.cpp', 'case WM_CTLCOLORSTATIC:',"return TRUE", '//')

comment_out_lines('windows/container.cpp', 'case WM_PAINT:',"return 0", '//')
comment_out_lines('windows/container.cpp', 'case WM_PRINTCLIENT:',"return 0", '//')


```

szanni:
I've recently been looking into this myself too. I'll check what your patches do. Apparently some people found some work arounds that I have been unable to reproduce so far.

Your code will probably trigger weird artifacts in some cases, at least that's what is suggested in the original issues you quoted. I have personally been unable to reproduce the supposed paint artifacts as well. Those might be related to certain theming?

e1732a364fed:
Just run the controls gallery example on windows, and when you resize the window, you'll see it's very laggy.

My patch is just a POC of a solution mentioned in the issue, to comment out some lines.

nptr:
I also encountered this issue during my work on a [libui-rs](https://github.com/libui-rs/libui) successor.
The changes proposed by the above script work well enough on my Win7 and Win10. The caveat is, that the tab page background is now "dialog gray" rather than "white", but it's worth it. No other artifacts related to this from what i can see. 
I assume issues arise when the tab page background isn't solid color or maybe even deviates from the default dialog color?

[...]

PR #180, opened 2023-03-27
Improving performance under Windows

nptr:
**Improving performance under Windows by dropping support for arbitrary backgrounds.**

I was asked to open a PR for my patch i offered in issue #161.

`libui` has a `container` control that is (seemingly) used to provide the basic layout mechanisms. It is supposed to be transparent.
Unfortunately GDI doesn't support real transparency and things get really hacky because of this. The Win32 controls way to achieve a similar effect is to ask the parent control for the appropriate background brush (WM_CTLCOLORSTATIC) or ask to the parent to render the background for it (WM_PRINTCLIENT).

The original author chose the second route, double buffered too. This works with every background (images, gradients etc.). It hurts performance pretty hard however.

_Note: I am not 100% sure if there isn't another reason why rendering is so slow. There seems to be some overdraw for example but I don't have the time to dig deeper._

The WM_CTLCOLORSTATIC route only supports solid color backgrounds. Since Microsoft doesn't use themed backgrounds since XP and most user themes also don't use gradients etc, the first approach seems to be a sufficient. This is what this PR is about. Performance is much better and there is no visual difference under the common Windows themes (Classic, Aero Basic, Aero Win7, Aero Win10). 

Please have a look and let me know if it works for you just as good. I can also post some screenshots in the next few days.
- nptr


nptr:
Here are the screenshots i offered. Solid backgrounds work fine and theme colors are respected.
### Win 7 Aero Theme
patch_w7_aero
## Win 7 Classic Theme
patch_w7_classic
## Win 7 High Contrast Theme
patch_w7_highcontrast
## Win 7 with Custom Theme
win7_themed

nptr:
Good catch! I'll see what's up with the label text. There is a chance I messed the WM_CTLCOLORSTATIC up.

nptr:
Seems like there is an expectation that the text color is set directly to the controls device context in the [WM_CTLCOLORSTATIC ](https://learn.microsoft.com/en-us/windows/win32/controls/wm-ctlcolorstatic) handler. MSDN doesn't really say that this is important but only portraits it as an option in my opinion.

I added the `SetTextColor(dc, GetSysColor(COLOR_WINDOWTEXT));` call and then squashed my commits, including a new commit message that more closely matches the existing ones.

cody271:
## Win11 Light Theme
image

windows/container.cpp
```diff
@@ -50,25 +69,46 @@ static LRESULT CALLBACK containerWndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LP
 		mmi->ptMinTrackSize.x = minwid;
 		mmi->ptMinTrackSize.y = minht;
 		return lResult;
+
+	// GDI doesn't support transparency. The Win32 controls way to achieve a similar effect
```
cody271, line 73: Nit: trailing whitespace here and below (GitHub UI doesn't detect this currently)
```
180.patch:57: trailing whitespace.
	// GDI doesn't support transparency. The Win32 controls way to achieve a similar effect 
180.patch:62: trailing whitespace.
	// other containers. So we have to dynamically find the next control that is not a 
warning: 2 lines add whitespace errors.
```

nptr: fixed it

windows/container.cpp
```diff
@@ -50,25 +69,46 @@ static LRESULT CALLBACK containerWndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LP
...
 		dc = BeginPaint(hwnd, &ps);
-		if (dc == NULL) {
-			logLastError(L"error beginning container paint");
-			// bail out; hope DefWindowProc() catches us
-			break;
-		}
-		r = ps.rcPaint;
-		paintContainerBackground(hwnd, dc, &r);
-		EndPaint(hwnd, &ps);
+		if(!dc) break;
```
szanni, line 88:
```suggestion
		if(dc == NULL)
			break;
```
For consistency

windows/container.cpp
```diff
@@ -14,6 +14,23 @@ struct containerInit {
...
+{
+	HWND parent;
+	int cls;
+
+	parent = hwnd;
+	for (;;) {
+		parent = parentOf(parent);
+		// skip groupboxes; they're (supposed to be) transparent
+		// skip uiContainers; they don't draw anything
+		cls = windowClassOf(parent, L"button", containerClass, NULL);
```
szanni, line 27: I know this is from existing code and doesn't seem to affect things but `L"button"` is used for a lot more than `uiGroup` and looks wrong to me. No need for action.

szanni (commented):
Looks very clean to me. More lines removed than added - that's how I like things.

I tested things on my Windows 7 VM. I did not notice any regressions. Painting is _a lot_ faster. Sadly still not instant, I guess that has to do with the delay of all the parent traversal and individual painting of controls?

Wine seems to work too.

A few things I noted that should probably be addressed in another PR but that are related and important enough to mention here:

- The `windowClassOf` looks kind of wrong to me, as `L"button"` is used for many controls besides `uiGroup`, like for example actual buttons. This seemingly doesn't seem to matter though? Which makes me wonder: how much of any of this code is actually needed at all?
- Another existing problem that seems to persist with these patches is the `uiCombobox` not getting drawn off the bat. Page4 and 16 highlight this quite nicely. It only happens when you first open the application and gets drawn once you move the mouse over it, resize the window, etc. 
screenshot

windows/container.cpp
nptr: Done. I squashed the changes with the previous whitespace fixes.

windows/container.cpp
nptr: That would be pushbuttons, radiobuttons, checkbox, commandlink and groupbox if I am not mistaken. Only groupboxes can be parents in libui and possibly in general. I also had to look at this code quite a bit, but its fine i think. Just Win32 common controls madness.

windows/container.cpp
nptr: Coding guidelines would be nice nice addition. Or even some code style checker or formatter like `clang-format`.
But i think that would require reformatting larger parts of the code base due to the (imo) unusual style choices in some areas.

nptr: About the `uiCombobox`: I was hoping it vanishes too but it didn't. Possibly a bad combination of window styles that result in bad ordering of the draw calls?

cody271: > About the `uiCombobox`: I was hoping it vanishes too but it didn't. Possibly a bad combination of window styles that result in bad ordering of the draw calls? On a second glance, it could also be a Z-ordering problem.

I've tracked this issue down somewhat.. nothing to do with `uiCombobox` specifcally.
Its an existing bug in `uiTab` [page](https://github.com/libui-ng/libui-ng/blob/master/windows/tabpage.cpp) control impl: tab pages are implemented as reparented dialogs on Windows.
See the old issue with test case to reproduce: https://github.com/andlabs/ui/issues/380#issuecomment-786319626

nptr: Anything left before this PR can be merged?

cody271: Nope I think its ready to go, thanks @nptr

cody271 (approved):
LGTM

PR #181, opened 2023-03-29
darwin: Fix uiEntryOnChanged callback not being triggered.

szanni:
The delegate object was released prematurely preventing the OnChanged() callback from being triggered.

Actual fallout can be seen in #133. This patch fixes things.


cody271 (approved):
LGTM

PR #184, opened 2023-03-29
darwin: Replace NSTextfield protocol with plain C functions.

szanni:
Protocols require the `-ObjC` linker flag when assembling binaries. This becomes burdensome when using libui-ng as a meson subproject.

Replacing the custom NSTextfield protocol with plain C functions solves this dilemma while having a slightly less fancy call signature.

Related to #152 - the existing protocols make #152 even worse. Removing them is a step in the right direction.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -76,8 +76,6 @@ if libui_OS == 'darwin'
 	libui_arch = ['-arch', 'x86_64', '-arch', 'arm64']
 	add_global_arguments(libui_arch, language: libui_darwin_langs)
 	add_global_link_arguments(libui_arch, language: libui_darwin_langs)
-
-	add_global_link_arguments(['-ObjC'], language: ['objc'])
 endif
 
 if libui_MSVC
--- a/darwin/entry.m
+++ b/darwin/entry.m
@@ -160,7 +160,7 @@ static NSTextField *realNewEditableTextField(Class class)
 	NSTextField *tf;
 
 	tf = [[class alloc] initWithFrame:NSZeroRect];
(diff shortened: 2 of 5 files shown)


meson.build
```diff
@@ -76,8 +76,6 @@ if libui_OS == 'darwin'
 	libui_arch = ['-arch', 'x86_64', '-arch', 'arm64']
 	add_global_arguments(libui_arch, language: libui_darwin_langs)
 	add_global_link_arguments(libui_arch, language: libui_darwin_langs)
-
-	add_global_link_arguments(['-ObjC'], language: ['objc'])
```
cody271, line 80: So this was also added to [fix](https://github.com/libui-ng/golang-ui/commit/2de08ed1b05864ad886dcda9fa71adfc30355a9b) linking in the Go bindings, should it be removed there as well?

szanni: Yes. This will not be needed anymore. I was not aware that _every_ project using libui-ng would need to add `-Objc` to their linker flags when adding protocols.
We currently do not have a pkgconfig file either to specify such flags and I am unclear how/if bindings really have awareness of pkgconfig in the first place.
Removing (and I guess generally banning) objective-c protocols and just writing things using free standing C functions is clean and removes any such concerns.

cody271: Confirmed working with this patch and no linker flag, Go bindings commit [reverted](https://github.com/libui-ng/golang-ui/commit/13dec0981460763493a936ddacc359d93422e827).

cody271 (approved):
LGTM

PR #185, opened 2023-03-30
Update list of bindings.

nptr:
I would like to introduce my libui-ng Rust binding to the binding list: [`libui`](https://github.com/libui-rs/libui)
_(For people knowing the original libui that is certainly confusing, for most in the Rust ecosystem it should not be :)_

Its actually a fork or continuation of [`iui`](https://github.com/rust-native-ui/libui-rs) (here listed as libui-rs), with the new features of libui-ng and of course a unique flavor to it. Like iui did before, libui-ng is also not compiled with the supplied build system (Meson + Ninja), but a custom build script calling the compilers directly to cut down on dependencies.

Please consider adding it to the list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ PHP | [ui](https://github.com/krakjoe/ui)
 Python | [pylibui](https://github.com/joaoventura/pylibui)
 Ring | [RingLibui](https://github.com/ring-lang/ring/tree/master/extensions/ringlibui)
 Ruby | [libui-ruby](https://github.com/jamescook/libui-ruby), [LibUI](https://github.com/kojix2/libui), [Glimmer DSL for LibUI](https://github.com/AndyObtiva/glimmer-dsl-libui)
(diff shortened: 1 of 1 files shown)


szanni:
Looks good, thanks for your contribution.

> libui-ng is also not compiled with the supplied build system (Meson + Ninja), but a custom build script calling the compilers directly to cut down on dependencies.

Obviously you figured out how to do things manually, but would cmake make your life easier? Or is that still too much of an dependency?

nptr:
The original authors did most of the work. I just happened to like it and plan to continue it as long as its reasonable.

CMake/Meson is not something a Rust user typically has, and there is already a build script running that generates the low level bindings, so why not invoke the compiler directly? For example, this gave me a hassle free installation on Win 7 because Meson 58 (as required by the meson build script) requires Win 10. The Rust build script just worked.

If I were to build something with C++, I likely have CMake already and the situation would be different :)

cody271 (approved):
LGTM

PR #187, opened 2023-03-31
Add API for getting and setting window position.

szanni:
This is the continuation of #62. I cleaned up the code and added support for darwin.

The current implementation does _not_ allow for `NULL` to be passed, mimicking the behavior in `uiWindowContentSize()`. IMO this might be useful to add at some point, but that can easily be done globally at a later stage without breaking anything.

To be aware: this is currently limited in usefulness as we do not have an API for getting screen dimensions. It however allows us to place windows in the top left corner or restoring the previous position of windows (think same window placement after an application restart).

As noted in the docs, this may not/only partially work on Unix platforms, similar to most of the other `uiWindow` stuff. GTK is very clear about this in the docs.

You can test in `tester` on `page15`.
```c
void uiWindowPosition(uiWindow *w, int *x, int *y);
void uiWindowSetPosition(uiWindow *w, int x, int y);
void uiWindowOnPositionChanged(uiWindow *w, void (*f)(uiWindow *sender, void *senderData), void *data);
```

Tested on macOS Big Sur, Windows 7, Linux Gtk3.24.37 and wine 8.3.

Closes #62

### DIFF
--- a/test/page15.c
+++ b/test/page15.c
@@ -210,6 +210,7 @@ void onMoved(uiWindow *w, void *data)
 {
 	int x, y;
 
+	puts("PositionChanged");
 	uiWindowPosition(w, &x, &y);
 	uiSpinboxSetValue(posX, x);
 	uiSpinboxSetValue(posY, y);
@@ -228,7 +229,7 @@ static void updatesize(uiWindow *w)
 
 void onSize(uiWindow *w, void *data)
 {
-	printf("size\n");
+	puts("ContentSizeChanged");
 	updatesize(w);
 }
 
--- a/test/page15.c
+++ b/test/page15.c
@@ -163,6 +163,7 @@ uiWindowSetChild(w,uiControl(b));}
 }
 
 static uiSpinbox *width, *height;
+static uiSpinbox *posX, *posY;
 static uiCheckbox *fullscreen;
 
 static void sizeWidth(uiSpinbox *s, void *data)
@@ -183,6 +184,35 @@ static void sizeHeight(uiSpinbox *s, void *data)
 	uiWindowContentSize(w, &xp, &yp);
 	yp = uiSpinboxValue(height);
 	uiWindowSetContentSize(w, xp, yp);
+}
+
+static void posXChanged(uiSpinbox *s, void *data)
+{
+	uiWindow *w = uiWindow(data);
+	int x, y;
+
+	uiWindowPosition(w, &x, &y);
+	x = uiSpinboxValue(posX);
+	uiWindowSetPosition(w, x, y);
+}
+
+static void posYChanged(uiSpinbox *s, void *data)
+{
+	uiWindow *w = uiWindow(data);
+	int x, y;
+
+	uiWindowPosition(w, &x, &y);
+	y = uiSpinboxValue(posY);
+	uiWindowSetPosition(w, x, y);
+}
+
+void onMoved(uiWindow *w, void *data)
+{
+	int x, y;
+
+	uiWindowPosition(w, &x, &y);
+	uiSpinboxSetValue(posX, x);
+	uiSpinboxSetValue(posY, y);
 }
 
 static void updatesize(uiWindow *w)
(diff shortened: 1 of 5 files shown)


test/page15.c
```diff
@@ -230,15 +260,30 @@ uiBox *makePage15(uiWindow *w)
...
+	uiBoxAppend(page15, uiControl(hbox), 0);
+
+	uiBoxAppend(hbox, uiControl(uiNewLabel("Position")), 0);
+	xpos = uiNewSpinbox(0, INT_MAX);
+	uiBoxAppend(hbox, uiControl(xpos), 1);
+	ypos = uiNewSpinbox(0, INT_MAX);
+
+	uiSpinboxOnChanged(xpos, posX, w);
+	uiSpinboxOnChanged(ypos, posY, w);
+	updatePos(w);
```
cody271, line 281: Patch to make the window position updates read/write instead of write only:
```patch
diff --git a/test/page15.c b/test/page15.c
index af35f084..167fff42 100644
--- a/test/page15.c
+++ b/test/page15.c
@@ -206,13 +206,16 @@ static void posY(uiSpinbox *s, void *data)
        uiWindowSetPosition(w, x, y);
 }
 
-static void updatePos(uiWindow *w)
+static int updatePos(void *data)
 {
        int x, y;
+       uiWindow *w;
 
+       w = uiWindow(data);
        uiWindowPosition(w, &x, &y);
        uiSpinboxSetValue(xpos, x);
        uiSpinboxSetValue(ypos, y);
+       return 1;
 }
 
 static void updatesize(uiWindow *w)
@@ -279,6 +282,7 @@ uiBox *makePage15(uiWindow *w)
        uiSpinboxOnChanged(xpos, posX, w);
        uiSpinboxOnChanged(ypos, posY, w);
        updatePos(w);
+       uiTimer(100, updatePos, w);
 
        uiBoxAppend(hbox, uiControl(ypos), 1);
        fullscreen = uiNewCheckbox("Fullscreen");
```

cody271:
Instead of having to poll for the current window position, shouldn't this API also support an event?
```c
_UI_EXTERN void uiWindowOnPositionChanged(uiWindow *w,
        void (*f)(uiWindow *sender, void *senderData), void *data);
```

szanni:
> Instead of having to poll for the current window position, shouldn't this API also support an event?

Yeah, it definitely should. Just haven't gotten around to it yet. I would very much prefer that to any timer based events.

test/page15.c
szanni: Solved via `uiWindowOnPositionChanged()`

szanni:
> Instead of having to poll for the current window position, shouldn't this API also support an event?

Just added your suggested event API. As initially feared, this is a little convoluted on unix as GTK does not really offer a proper API to differentiate between move and resize events. I solved this by just caching and comparing the values.

Note: I found the GTK gem in `old/_abort/windowevents/unix_window.c`. The matching darwin implementation actually seems broken...  I guess all of this used to be part of the API at some point? From `old/news.md`:

> Due to being unable to guarantee they will work (especially as we move toward capability-driven window systems like Wayland), or being unable to work without hacking that breaks other things, the following functions have been removed: `uiWindowPosition()`, `uiWindowSetPosition()`, `uiWindowCenter()`, and `uiWindowOnPositionChanged()`.

I disagree with the reasoning for removal. I believe the best effort approach and a note in the API docs should suffice.

I do have patches for the unix `ContentsizeChanged` insanity in the pipeline as well. I will commit once this gets merged to avoid merge conflicts.

cody271:
> I disagree with the reasoning for removal. I believe the best effort approach and a note in the API docs should suffice.
>
>
Yes of course, "best effort" is perfectly acceptable and necessary here.

> I do have patches for the unix `ContentsizeChanged` insanity in the pipeline as well. I will commit once this gets merged to avoid merge conflicts.
>
👍

test/page15.c
```diff
@@ -198,7 +229,7 @@ static void updatesize(uiWindow *w)
 
 void onSize(uiWindow *w, void *data)
 {
-	printf("size\n");
+	puts("ConentSizeChanged");
```
cody271, line 232:
```suggestion
	puts("ContentSizeChanged");
```

szanni: Fixed. Good catch.

szanni:
Fixed the typo and added (IMO) better variable names for the tester.

cody271 (approved):
LGTM

PR #189, opened 2023-04-05
unix: Fix uiWindowContentSizeChanged()

szanni:
This fixes `uiWindowOnContentsizeChanged()` getting triggered randomly on mouse moves as well as masking the callback on `uiWindowSetContentSize()` to match darwin, windows and other API calls.

Updated docs & removed `@todo`.


cody271 (approved):
LGTM

PR #194, opened 2023-04-08
darwin: Fix uiWindow fullscreen/borderless/resize interaction.

### LINKED ISSUE #190: darwin: uiWindowResizeable() broken with uiWindowFullscreen() and/or uiWindowBorderless()
uiWindowSetFullscreen():
In `tester` uncheck `Menu > Test > Enable Resize`. Go to page15 and check `Fullscreen`. Nothing happens. Both unix and 
windows go into full screen mode, which is what I expect.

uiWindowResizable():
In `tester` uncheck `Menu > Test > Enable Resize`. Go to page15 and check `Borderless`. Uncheck `Borderless` and try to resize the window. The window is suddenly resizable again although it should not be.

The entirety of uiWindow on darwin is a hot mess and IMO needs to be rewritten. 

Tested on macOS Big Sur.

Related to #186

Edit: Also reproducible via a patched `qa` as suggested in #192

- QA: 80f6d6d > darwin (macOS 11.7.1) > uiWindow > 4. Fullscreen + Borderless > Step 3: Window does not enter full screen when borderless
- QA: 80f6d6d > darwin (macOS 11.7.1) > uiWindow > 5. Fullscreen + Resizeable > Step 6: Window does not exit full screen when not resizeable

szanni:
- Move window display settings to a separate function updateStyleMask().
- Determine full screen by querying the window styleMask.

Fix:
- Window not going into fullscreen when borderless.
- Window not retaining not resizeable when toggling borderless.
- Window not going into fullscreen when not resizeable.

Fixes #190 for me.

Two things to mention:
1. Transitioning from borderless to fullscreen will set borders before transitioning into fullscreen which is visible on my system for a split second. I did not find any way around this.
2. Resize is kept enabled in fullscreen, so that the user can exit fullscreen via the window controls. Querying `uiWindowResizeable()` will still return `false` and the correct state is restored when exiting fullscreen.


szanni:
Updated `updateStyleMask()` to be static and used cast to `BOOL` instead of `!!`.

cody271:
> 2\. Resize is kept enabled in fullscreen, so that the user can exit fullscreen via the window controls.

In testing the latest here, I noticed that this suffers from a [similiar issue](https://github.com/libui-ng/libui-ng/pull/187#discussion_r1155067369) as window position `page15` test had.
Exiting fullscreen via window controls means the checkbox is then out of sync with the fullscreen state, and the only way to stay in sync is either through polling the state or an event.
Maybe we need to add `uiWindowOnFullscreenChanged()` like we did for `uiWindowOnPositionChanged()` ?

szanni:
> In testing the latest here, I noticed that this suffers from a [similiar issue](https://github.com/libui-ng/libui-ng/pull/187#discussion_r1155067369) as window position `page15` test had. Exiting fullscreen via window controls means the checkbox is then out of sync with the fullscreen state, and the only way to stay in sync is either through polling the state or an event. Maybe we need to add `uiWindowOnFullscreenChanged()` like we did for `uiWindowOnPositionChanged()` ?

That is correct and while `uiWindowFullscreen()` does return the correct state, I would love to have a `uiWindowOnFullscreenChanged()` and some `qa` instructions for darwin to test full screen via the window decorations.

Issue #191, opened 2023-04-06
API string ownership clarification


bcampbell:
From the tooltip api issue:
```c
/**
 * Sets the control tooltip.
 *
 * @param c uiControl instance.
 * @param tooltip Control tooltip.\n
 *             A valid, `NULL` terminated UTF-8 string.\n
 *             Data is owned by the caller.
 * @note Setting `NULL` resets the tooltip to the default value.
 * @memberof uiControl
 */
_UI_EXTERN void uiControlSetTooltip(uiControl *c, const char *tooltip);
```

_Originally posted by @cody271 in https://github.com/libui-ng/libui-ng/issues/188#issuecomment-1496781878_

My reading of that `uiControlSetTooltip()` doc comment is that the caller needs to ensure that the string pointed to by `tooltip` exists for the lifetime of the control, right?

I wanted to check that this was the same convention used elsewhere in libui. I picked uiButton as a random example:
And sure enough, `uiSetButtonText()` uses the same wording: "Data is owned by the caller".
But `uiButtonText()` says of the returned sting: "Data is owned by the caller, make sure to call `uiFreeText()`".

That seems rather... asymmetrical to me.
Is this really the general convention or am I just misreading the "Data is owned by the caller." comment?
- Set functions just stash a pointer without copying the string (and rely on the caller to ensure the string stays valid!)
- Get functions return a copy which must be freed by the caller

bcampbell:
The more I think about it, the more I think I'm just misreading the comment. The API _must_ be taking a copy of the passed-in string.
Peeking through the code, the GTK and win32 `uiButton` implementations leave the string-copy policy up to those APIs.
And, while [`gtk_button_set_label()`](https://developer-old.gnome.org/gtk3/stable/GtkButton.html#gtk-button-set-label) and [`SetWindowTextW()`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowtextw) don't say so explicitly, I can't imagine that they _don't_ take a copy. Anything else would be madness.

So. Would anyone be open to a patch to change the wording on those comments in `ui.h`?
Out with: "Data is owned by the caller."
In with "The control will make it's own copy of the string." (or something less unwieldly...)

szanni:
Yeah, my phrasing may have been a little confusing here.
`Data is owned by the caller.` meaning it is not transferred and ownership and memory management stays with the caller. Or as you say  `The control will make a copy.`.
[...]

PR #195, opened 2023-04-09
Clarify string ownership documentation (issue #191)

bcampbell:


cody271 (commented):
Looks good, but I still count 13 lines with "Data is owned by the caller" in `ui.h`:
```shell
libui-ng$ grep caller ui.h | wc -l
13
```

bcampbell:
> Looks good, but I still count 13 lines with "Data is owned by the caller" in `ui.h`:

Those ones are all where data is being returned, and it isn't so ambiguous eg:

`Data is owned by the caller, make sure to call `uiFreeText()`

I think the `uiFreeText()` mention makes it clear enough. The "Data is owned by caller" scans a little oddly, but is technically correct... open to suggestions!

cody271:
> The "Data is owned by caller" scans a little oddly, but is technically correct... open to suggestions!

Yes I think this last part should be changed, since even I was confused by that language from a quick glance at it .. 😛 

Maybe just `Make sure to call uiFreeText().` ?

cody271:
> Those ones are all where data is being returned

Also in these cases maybe the text should be in a `@note`, so that they're harder to miss?

szanni:
For `in` I like the phrasing: `Data is copied internally. Ownership is not transferred.`
I would personally not differentiate between `String` and `Data` and just keep the phrasing consistent for all calls.

For the `out` (return) maybe something like: `Data is copied and ownership is transferred to the caller. Make sure to call uiFreeText().`

The other phrasing I like would be using `clone`:
`Data is cloned internally. Ownership is not transferred.`
`Data is cloned and ownership is transferred to the caller. Make sure to call uiFreeText().`


> > Those ones are all where data is being returned
> 
> Also in these cases maybe the text should be in a `@note`, so that they're harder to miss?


I don't believe moving things to a `@note` is needed or necessarily helpful, as the notes get rendered at the bottom and not where the return is described.  When mixing this with parameters this will get confusing again, as it will be unclear what it refers to exactly. Keeping it in visual proximity is the way to go IMO.
In general this should mainly be regarded as a hint for people that did not read the object lifetimes section of the manual (to be written). We should have a general section describing that: memory in is copied; memory out is copied and you need to free it.
In calls where this is not the case we should probably have a `@warning`.

bcampbell:
OK - new version with some rewording...

cody271 (approved):
LGTM

PR #197, opened 2023-04-10
windows: Fix rendering of some controls when inside uiTab pages

cody271:
Initial "fix" for the rendering artifacts in controls with dropdown content such as comboboxes: ~~Trigger a redraw of the tab page children on the next immediate frame draw after each window configure event of the tab control.~~

~~This introduces a small flicker on aggressive resizing of the window, but shouldn't notice any other performance issues like before. So that is an improvement over some controls just randomly missing from the tab page..~~

See the old issue with test case to reproduce: https://github.com/andlabs/ui/issues/380#issuecomment-786319626
Also `controlgallery` -> `Numbers and Lists`

### DIFF
--- a/windows/tab.cpp
+++ b/windows/tab.cpp
@@ -38,6 +38,8 @@ static void tabRelayout(uiTab *t)
 {
 	struct tabPage *page;
 	RECT r;
+	LONG_PTR controlID;
+	HWND insertAfter;
 
 	// first move the tab control itself
 	uiWindowsEnsureGetClientRect(t->hwnd, &r);
@@ -48,7 +50,10 @@ static void tabRelayout(uiTab *t)
 		return;
 	page = tabPage(t, curpage(t));
 	tabPageRect(t, &r);
+	controlID = 100;
+	insertAfter = NULL;
 	uiWindowsEnsureMoveWindowDuringResize(page->hwnd, r.left, r.top, r.right - r.left, r.bottom - r.top);
+	uiWindowsEnsureAssignControlIDZOrder(page->hwnd, &controlID, &insertAfter);
 }
 
 static void showHidePage(uiTab *t, LRESULT which, int hide)


nptr:
Did you investigate whether it might be a Z-ordering issue? During my performance fix i stumbled across [this post](https://stackoverflow.com/questions/27085427/common-controls-are-not-properly-painted-when-i-resize-window/58459674#58459674) and it seemed a good fit. Unfortunately I didn't find the time to properly investigate this route.

My plan was to print the Z-order somehow (Spy++ couldn't tell me Z-order?) and see if this might indeed be the problem.

cody271:
> Did you investigate whether it might be a Z-ordering issue?
>
Just pushed a much simpler Z-order based fix with `uiWindowsEnsureAssignControlIDZOrder()` that seems to work!
(Tab pages also use `uiWindowsControlAssignSoleControlIDZOrder()`, which _doesn't_ work..)

nptr:
Very nice! I hope to try it today or tomorrow 👍

nptr:
I can confirm that the combo boxes inside the tabs don't vanish anymore. Although now I believe the whole tab control is flickering.
Before, only the active tab _occasionally_ flickered. On the grand scale it's a win though and the direction seems to be right.

cody271:
> Although now I believe the whole tab control is flickering.

This is with the very latest patch? [First attempt](https://github.com/libui-ng/libui-ng/commit/cd8f3e6229f319fd5f1a05348fbbecdbfe794f5d) did still have some flickering, but that should be gone now with the [Z-order based](https://github.com/libui-ng/libui-ng/commit/12f750f92439e5d73c7a2acdb0af9447dc0a5794) fix.

cody271:
> Yes, the one with the Z-order fix. Tested on Win7 with the classic theme on.

I can't reproduce that on modern Windows. Curious then if my _first_ fix is actually better for legacy Windows?
If needed.. we could actually incorporate both of these hacks, and pick the best one at runtime based on Windows version.

nptr:
I'll test on my Win 10 machines too. Personally I'd prefer the z-order fix, even with the flicker. It might not be ideal or even the real solution but is an improvement nonetheless.

msink:
Tested on Win7 (mingw64 version):
- current version - page content does'nt blink on resize, but area with tab labels blinks,
- first vesrion - page content blinks on resize, area with tab labels doesn't.

nptr:
Indeed no resize-flicker on Win10 anymore. Even if starting the applications in compatibility mode for Win7. Z-Order is also the same, so no silent fixes from the OS. I only see a flicker on first tab change in the control gallery.

Pretty sure its still not how it is supposed to be, but probably good enough for the time being.

nptr:
Ok, i might have found our answer: The flickering tab is just how it works. I created a C# WinForms application with a tab control and it **flickers out of the box**! Meaning the engineers at MS couldn't even do better. And they don't even fully use the native tab control ("SysTabControl32") but made a good portion anew to handle the tab page creation better (no more dialog handling!).

Searching for tab control issues in the context of C# got me to [this question on stackoverflow](https://stackoverflow.com/questions/16925369/tabcontrol-flickering) and confirmed my suspicion from the paragraph above.

I also tried setting WS_EX_COMPOSITED on the main window (as mentioned in the stackoverflow question) and it indeed removed the flickering. Now there are issues in the "tester" however. Some tab pages don't render on start but only after a resize. So there really is no solution to the tab flickering.

nptr:
@cody271 Are you still working on the fix, require more tests, or happy with it and ready to merge?

PR #83, opened 2022-05-04
MinGW: compiling on clang

msink:

PR #198, opened 2023-04-24
Add build.zig

desttinghim:
These commits add a `build.zig` to libui. This allows for cross-compilation using the zig toolchain (compiling for windows from linux is as easy as `zig build -Dtarget=x86_64-windows`) and for using zig's built-in package manager (currently only in the nightly build) to directly depend on libui-ng. 

The build.zig has not been configured to build for MacOS, though this should be easy to add.

It might be best to hold off on merging this until there is a stable release of zig including the package manager. The current expected release date is May. 

I'm happy accepting constructive criticism on this PR; if you don't feel like merging the build.zig, that's fine - I mostly wrote it for a private project, and I am making a PR just in case it would be appreciated :)


cody271:
> These commits add a `build.zig` to libui. This allows for cross-compilation using the zig toolchain
>
Hi @desttinghim, this looks neat!

I'd be particularly interested in using this to cross-compile the [Go bindings](https://github.com/libui-ng/golang-ui). Is this possible?

> The build.zig has not been configured to build for MacOS, though this should be easy to add.
> 
Yes support for all three platforms would be required to merge this.

build.zig
```diff
@@ -0,0 +1,216 @@
...
+            "windows/tablemetrics.cpp",
+            "windows/tabpage.cpp",
+            "windows/text.cpp",
+            "windows/utf16.cpp",
+            "windows/utilwin.cpp",
+            "windows/window.cpp",
+            "windows/winpublic.cpp",
+            "windows/winutil.cpp",
+        }, &.{});
+    } else if (target.isLinux()) {
```
cody271, line 117: This breaks non-Linux platforms. The detection logic should be something like:
```zig
if (target.isWindows()) {
  // use windows/*.cpp backend
} else if (target.isDarwin()) {
  // use darwin/*.m backend
} else {
  // assume unix/*.c backend
}
```

desttinghim:
> > These commits add a `build.zig` to libui. This allows for cross-compilation using the zig toolchain
> 
> Hi @desttinghim, this looks neat!
> 
> I'd be particularly interested in using this to cross-compile the [Go bindings](https://github.com/libui-ng/golang-ui). Is this possible?

Yes, I believe so. I have no experience with using go myself, but there have been some posts on using `zig cc` for cross-compiling cgo binaries: https://zig.news/kristoff/building-sqlite-with-cgo-for-every-os-4cic

> > The build.zig has not been configured to build for MacOS, though this should be easy to add.
> 
> Yes support for all three platforms would be required to merge this.

This makes sense. I don't have any Apple devices I can use to test, but I'll push a `build.zig` that should theoretically work, if someone would like to test it.

cody271:
This PR produced the following error when I tried testing, same error on both Mac and Linux:
```shell
$ zig build
build.zig:249:18: error: expected ')', found ','
    for (examples, example_names) |example, name| {
                 ^
```
And if I try removing the code section for the examples, I then get this error:
```shell
$ zig build
build.zig:3:21: error: root struct of file 'std' has no member named 'Build'
pub fn build(b: *std.Build) void {
                 ~~~^~~~~~
```

cody271:
> and for using zig's built-in package manager (currently only in the nightly build) to directly depend on libui-ng.

Okay I'm getting past the above `build.zig` compile errors by using the nightly build of Zig.
So this build file doesn't appear to work against the latest stable release of Zig at all?

Build still fails though: Now seeing _lots_ of Objective-C runtime related link errors on Mac..

desttinghim:
Yes, this is using zig nightly. There have been a lot of changes since the latest stable release. I plan to stick to using zig 0.11 when it is released to make it easier for others to follow. 

Could you post the errors you are seeing?

build.zig
```diff
@@ -0,0 +1,275 @@
...
+        "common/table.c",
+        "common/tablemodel.c",
+        "common/tablevalue.c",
+        "common/userbugs.c",
+        "common/utf.c",
+    }, &.{});
+
+    if (target.isDarwin()) {
+        // use darwin/*.m backend
+        lib.addFrameworkPath("Foundation");
```
cody271, line 35:
```suggestion
        lib.linkFramework("Foundation");
```

build.zig
```diff
@@ -0,0 +1,275 @@
...
+        lib.addFrameworkPath("Appkit");
```
cody271, line 36:
```suggestion
        lib.linkFramework("Appkit");
```

cody271 (commented):
I think I fixed it:

cody271:
> I mostly wrote it for a private project, and I am making a PR just in case it would be appreciated :)

Curious are you also working on Zig bindings, or using the C API directly from Zig?
Would be great to have Zig bindings to add to [the list](https://github.com/libui-ng/libui-ng#language-bindings).. ;)

desttinghim:
I am using the C API directly at the moment. It wouldn't be terribly difficult to wrap into a more ziggish API but wrapping it is not a priority for me at the moment :) that being said, I wouldn't mind working on a set of bindings when I have the time.

cody271:
> that being said, I wouldn't mind working on a set of bindings when I have the time.

Even just our [`hello-world`](https://github.com/libui-ng/libui-ng/blob/master/examples/hello-world/main.c) example in Zig (with the C API directly and an example build file) would be quite helpful to those trying to get started with a new app in Zig.

build.zig
```diff
@@ -0,0 +1,275 @@
...
+        "common/shouldquit.c",
+        "common/table.c",
+        "common/tablemodel.c",
+        "common/tablevalue.c",
+        "common/userbugs.c",
+        "common/utf.c",
+    }, &.{});
+
+    if (target.isDarwin()) {
+        // use darwin/*.m backend
```
cody271, line 34: Missing `installHeader()` and `addIncludePath()` here?

build.zig
```diff
@@ -0,0 +1,275 @@
...
+        lib.linkSystemLibrary("msimg32");
+        lib.linkSystemLibrary("comdlg32");
+        lib.linkSystemLibrary("d2d1");
+        lib.linkSystemLibrary("dwrite");
+        lib.linkSystemLibrary("ole32");
+        lib.linkSystemLibrary("oleaut32");
+        lib.linkSystemLibrary("oleacc");
+        lib.linkSystemLibrary("uuid");
+        lib.linkSystemLibrary("windowscodecs");
+        lib.defineCMacro("ZIG", "1");
```
cody271, line 109: While it's not a big deal if so, I'm surprised to see we need our own custom define here.

Does the Zig compiler really not have a builtin define for detection like every other toolchain?

TUSF: Zig's C/C++/ObjC compiler is basically clang embedded, so they probably haven't seen a need to add another builtin define on top of Clang's.

cody271: Does existing `WIDL_EXPLICIT_AGGREGATE_RETURNS` define work here instead? I think its the same bug, see #211.

desttinghim: That does work! I've modified the files to use the `WIDL_EXPLICIT_AGGREGATE_RETURNS` macro and removed the ZIG macro.

desttinghim:
@cody271now that zig 0.11.0 has been released, I am interested in getting this into a mergeable state. The 0.11.0 release is the first stable version of zig to include a package manager. Using the nightlies would be difficult due to the frequent changes - pinning to a stable release makes it much easier to provide support.

windows/winutil.cpp
```diff
@@ -140,7 +140,7 @@ void invalidateRect(HWND hwnd, RECT *r, BOOL erase)
 // that damn ABI bug is never going to escape me is it
 D2D1_SIZE_F realGetSize(ID2D1RenderTarget *rt)
 {
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || defined(WIDL_EXPLICIT_AGGREGATE_RETURNS)
```
cody271, line 143: Have you tried using just `__clang__` here with Zig? See [#83](https://github.com/libui-ng/libui-ng/pull/83/files#diff-c71e43c3363ceaf0e49297ac212dab616e252ec4760a5e85b38ed2fbba310646)

desttinghim: I think I might have right at the beginning - it has been too long for me to remember clearly. I think this macro is better as it targets the feature/bug more specifically, but I will give this a try when I am back at my computer.

desttinghim: Ok, I've given the `__clang__` macro a try and it appears to work. Not sure which is the better macro to use

README.md
```diff
@@ -68,6 +68,30 @@ For the sake of completeness, I should note that the default value of `--layout`
...
+```
+
+#### Requirements
+On all hosts, [zig 0.11.0](https://ziglang.org/download/) must be installed.
+
+To compile for Linux, `gtk+-3.0` and `pkg-config` must be installed.
+
+To compile for Windows, a Windows resource compiler must be installed. The build.zig will default to `rc.exe` on Windows and `wrc` on non-Windows systems. This can be overriden by passing `-Dresource-compiler=<path to resource compiler executable>`.
+
+To compile for MacOS, you will need the MacOS sdk installed. As far as I know, this requires an Apple device.
```
cody271, line 93:
```suggestion
Cross compilation for macOS targets from other platforms is not supported at this time. Apple macOS SDK is only licensed for use from Apple devices.
```

cody271 (commented):
Few more nits / suggestions

cody271:
Is this still intended to work with the recent Zig stable release, or does it need dev nightly?

Packages for `0.11.0` have now been released for most systems.

desttinghim:
The new commits I pushed are not for 0.11.0, unfortunately. I updated to a nightly version of zig while troubleshooting a libui project that was failing to run on windows. The nightly version integrates a windows resource compiler into zig which simplifies getting a toolchain set up for windows significantly. I can move the most recent commit to a new branch and roll back to 0.11.0 here if you'd like.

cody271:
> I can move the most recent commit to a new branch and roll back to 0.11.0 here if you'd like.

Ideally, this would support _both_ stable from packages and dev nightly with the same build script.

Could check `builtin.zig_version` first to decide which resource compiler to use?

cody271:
> I updated to a nightly version of zig while troubleshooting a libui project that was failing to run on windows.
> The nightly version integrates a windows resource compiler into zig which simplifies getting a toolchain set up for windows significantly.

After having actually tried running this on a new machine, changed my mind: you had it right the first time. 😛 
Requiring a third party resource compiler is very painful.. using the builtin Zig resource compiler is a much better idea.
As long as the docs are updated to require `0.12.0`, I think its okay.

desttinghim:
Ok, this is back to the build.zig for 0.12.0. Hopefully there aren't too many breaking changes before the release :sweat_smile: If we want to guarantee support for the latest nightly at all times, we will probably want to set up CI to download the latest version and run `zig build`. I know other zig projects have something like that set up, though I haven't personally done it. Otherwise we can just pin to a known working version and update when a stable version is released.

cody271:
> Ok, this is back to the build.zig for 0.12.0.
>
> Otherwise we can just pin to a known working version and update when a stable version is released.

Yes the hope is that the upcoming 0.12.0 stable release from packages will just work, without depending on nightly download..

desttinghim:
Another useful update has been merged into zig: https://github.com/ziglang/zig/pull/18160

The short version is that zig modules with a c dependency will now forward that dependency. Previously all the c dependencies had to be added for each module you wanted to depend on it.

I will update this branch to make use of the new build feature update the documentation on which zig version to use.

On a related note, should I try and add the zig build step to the CI?

PR #199, opened 2023-04-30
unix: Fix uiMsgBoxError()

cody271:
Recent Gtk3 versions ignore the hint `GTK_MESSAGE_ERROR`, and render all message dialogs the same now:

This PR restores the old behavior:

### DIFF
--- a/unix/stddialogs.c
+++ b/unix/stddialogs.c
@@ -56,6 +56,11 @@ static void msgbox(GtkWindow *parent, const char *title, const char *description
 		type, buttons,
 		"%s", title);
 	gtk_message_dialog_format_secondary_text(GTK_MESSAGE_DIALOG(md), "%s", description);
+	if (type == GTK_MESSAGE_ERROR) {
+		gtk_message_dialog_set_image(GTK_MESSAGE_DIALOG(md),
+			gtk_image_new_from_icon_name("dialog-warning", GTK_ICON_SIZE_DIALOG));
+		gtk_widget_show_all(md);
+	}
 	gtk_dialog_run(GTK_DIALOG(md));
 	gtk_widget_destroy(md);
 }


unix/stddialogs.c
```diff
@@ -56,6 +56,11 @@ static void msgbox(GtkWindow *parent, const char *title, const char *description
 		type, buttons,
 		"%s", title);
 	gtk_message_dialog_format_secondary_text(GTK_MESSAGE_DIALOG(md), "%s", description);
+	if (type == GTK_MESSAGE_ERROR) {
+		gtk_message_dialog_set_image(GTK_MESSAGE_DIALOG(md),
+			gtk_image_new_from_icon_name("dialog-warning", GTK_ICON_SIZE_DIALOG));
```
szanni, line 61:
```suggestion
			gtk_image_new_from_icon_name("dialog-error", GTK_ICON_SIZE_DIALOG));
```
Should this not be `dialog-error`?
Not sure what older GTK versions used to display, but I would expect an error symbol for uiMsgBoxError.

szanni (commented):
Looks good apart from the warning/error image. I presume the code would simply override the stock image on older Gtk3 versions?

unix/stddialogs.c
cody271: > Looks good apart from the warning/error image. I presume the code would simply override the stock image on older Gtk3 versions?

Yes will just override in that case.

> Should this not be dialog-error?
> Not sure what older GTK versions used to display, but I would expect an error symbol for uiMsgBoxError.

Intentionally used `dialog-warning` instead of `dialog-error` here, as the Gtk stock icon for `dialog-error` maps to a somewhat misleading "stop sign" symbol instead of an "error" symbol:



`dialog-warning` matches the other platforms best as well:

darwin_errmsg

szanni: > `dialog-warning` matches the other platforms best as well:

Interesting that that is the way `NSCriticalAlert` is displayed on darwin. Maybe not to scare people?

This is what the dialog looks like on Windows 7, not sure about newer platforms:
screenshot

I guess having both a warning and  an error box would make sense?

cody271: > not sure about newer platforms

Here it is on Windows 11, nearly identical:

windows_msgboxerr


> I guess having both a warning and an error box would make sense?

Yes, `uiMsgBoxWarning()` would be a good addition to the API as well.

PR #206, opened 2023-05-17
Floating point spinbox

### LINKED ISSUE #201: Floating Point Spinbox
I have need for a numerical input, like a spinbox, but for floating point numbers. I've seen this in wxWidgets, they call it a DoubleSpinBox.

desttinghim:
This changes the spinbox control to use a `double` value. The integer limitation from before was mostly due to windows up-down control not natively handling floating point. This is a breaking API change for spinbox related functions.

Closes #201 

# Platforms
- [x] Unix
- [x] Windows
- [ ] Darwin


desttinghim:
I've made an attempt at writing the changes for darwin. I don't have a mac machine so I can't test the changes, but hopefully it will be a good jumping off point for someone else.

cody271 (commented):
This should be its own new control type `uiSpinboxDouble`, as mentioned in the issue.
Also failing CI. Otherwise looks good so far!

desttinghim:
Will do - I wasn't sure whether changing the existing control or adding a new one would be better. It looks like the MacOS changes didn't work; not terribly surprising since I'm working blind.

common/controlsigs.h
```diff
@@ -22,6 +22,7 @@
 #define uiSeparatorSignature 0x53657061
 #define uiSliderSignature 0x536C6964
 #define uiSpinboxSignature 0x5370696E
+#define uiSpinboxDoubleSignature 0x5370696E
```
cody271, line 25:
```suggestion
#define uiSpinboxDoubleSignature 0x53706E44
```
If you add a new control, each control type needs to be assigned a unique type ID value.
(The type IDs are actually ASCII to assist in debugging: `0x5370696E` = "Spin", `0x53706E44` = "SpnD".)

darwin/spinbox.m
```diff
@@ -123,58 +123,58 @@ - (void)dealloc
...
+	[self libui_setValue:[self->tf value]];
 	(*(self->spinbox->onChanged))(self->spinbox, self->spinbox->onChangedData);
 }
 
 @end
 
 uiDarwinControlAllDefaults(uiSpinbox, spinbox)
 
-int uiSpinboxValue(uiSpinbox *s)
+double uiSpinboxValue(uiSpinbox *s)
```
cody271, line 172: This file still has the API breaking changes. Fine to just stub out the Darwin implementation for now.

examples/histogram/main.c
```diff
@@ -1,9 +1,9 @@
 // 13 october 2015
```
cody271, line 1: Revert all this. There should be no changes needed to this file in this PR.

unix/spinboxdouble.c
```diff
@@ -0,0 +1,77 @@
+// 11 june 2015
```
cody271, line 1: Instead of copy pasting the entire thing, the implementation should be parameterized internally to support both integer and double modes. See `uiEntry` implementations for an example of this pattern.

darwin/spinbox.m
desttinghim: I've stubbed it out in the latest changes I've pushed

desttinghim:
Ok, I think the code is almost there. The commit history is a little messy, should I squash my commits and force push?

unix/spinboxdouble.c
cody271: > If I understand correctly, this would mean changing the functions to operate on a uiSpinbox and removing the new uiSpinboxDouble type? That seems reasonable, but I'm just double checking that I understand what you mean correctly.

Yup exactly, adding new "double mode" for existing `uiSpinbox` control type should work here.

cody271:
> Ok, I think the code is almost there. The commit history is a little messy, should I squash my commits and force push?

Yes please try to clean up the commit history as best you can before merge.

desttinghim:
I think this branch is ready for merging as soon as the Darwin backend gets implemented. Wish I could be more helpful on that front!

cody271:
This should throw an error if `uiSpinboxSetValueDouble()` is called on an integer mode `uiSpinbox`.

desttinghim:
To me it seems valid to allow getting the value as a double even in int mode. I don't see a good reason to disallow it.

cody271:
> This should throw an error if uiSpinboxSetValueDouble() is called on an integer mode uiSpinbox.
>
>> To me it seems valid to allow getting the value as a double even in int mode. I don't see a good reason to disallow it.

Yes _getting_ the value as double in int mode is safe, but setting it is the problem.
Double to int conversion silently fails with precision loss, but the reverse does not.

desttinghim:
I didn't read closely enough the first time! I've added a check.

desttinghim:
Hey, is this ready to go? I'd love to get this merged

cody271:
> Hey, is this ready to go? I'd love to get this merged

Its fairly close! I'll work on the Darwin implementation for this next, it should be simple.

matyalatte:
Hi, I really wanted this feature for my app.
So, I made the Darwin version by myself.
You can see three commits about the float picker on my branch.
- [Add float picker for Darwin](https://github.com/matyalatte/libui-ng/commit/2007d3718168dc0b172328a344aa16f977acc78c)
- [Tiny improvements for Windows and Unix](https://github.com/matyalatte/libui-ng/commit/d14ca59694b2fd2669a986e48c40f180e8e80565)
- [Add `uiSpinboxValueText` to get the formatted text](https://github.com/matyalatte/libui-ng/commit/a22b20c42d7f37961fcb63a859f83bedb69fc4f1)

I think all of them can be applied to this PR.

desttinghim:
Hey @matyalatte, sorry for the long delay! I added your changes to this branch. Thank you for getting the Darwin version implemented :slightly_smiling_face:

PR #207, opened 2023-05-18
file/folder dialog parameters: path, name, filters

### LINKED ISSUE #202: Allow suggesting filename with uiSaveFile
I don't know if this is a standard feature on every os, but I'd like to be able to suggest a filename for the user when they save something with uiSaveFile. If it's supported with every GUI toolkit, I think it would make sense to change the uiSaveFile function to accept a pointer to an options struct. If null is passed, it will use the default values.

desttinghim:
This is revival of #47, since it appears the original author has not been active on Github for some time now. Closes #202

- [x] Fix for null filter string
- [x] Linux support
- [x] Windows support
- [ ] Mac support

### DIFF
--- a/darwin/stddialogs.m
+++ b/darwin/stddialogs.m
@@ -73,6 +73,8 @@ char *uiSaveFile(uiWindow *parent)
 char *uiOpenFileWithParams(uiWindow *parent, uiFileDialogParams *params)
 {
 	// TODO
+	uiprivImplBug("Not yet implemented.");
+	return NULL;
 }
 
 char *uiOpenFolderWithParams(uiWindow *parent, uiFileDialogParams *params)
--- a/darwin/stddialogs.m
+++ b/darwin/stddialogs.m
@@ -32,7 +32,7 @@ static char *runSavePanel(NSWindow *parent, NSSavePanel *s)
 	return filename;
 }
 
-char *uiOpenFile(uiWindow *parent, const char *defaultPath, const char *filter)
+char *uiOpenFile(uiWindow *parent)
 {
 	NSOpenPanel *o;
 
@@ -46,7 +46,7 @@ char *uiOpenFile(uiWindow *parent, const char *defaultPath, const char *filter)
 	return runSavePanel(windowWindow(parent), o);
 }
 
-char *uiOpenFolder(uiWindow *parent, const char *defaultPath)
+char *uiOpenFolder(uiWindow *parent)
 {
 	NSOpenPanel *o;
 
@@ -60,7 +60,7 @@ char *uiOpenFolder(uiWindow *parent, const char *defaultPath)
 	return runSavePanel(windowWindow(parent), o);
 }
 
-char *uiSaveFile(uiWindow *parent, const char *defaultPath, const char *defaultName, const char *filter)
+char *uiSaveFile(uiWindow *parent)
 {
 	NSSavePanel *s;
 
@@ -68,6 +68,21 @@ char *uiSaveFile(uiWindow *parent, const char *defaultPath, const char *defaultN
 	setupSavePanel(s);
 	// panel is autoreleased
 	return runSavePanel(windowWindow(parent), s);
+}
+
+char *uiOpenFileWithParams(uiWindow *parent, uiFileDialogParams *params)
+{
+	// TODO
+}
+
+char *uiOpenFolderWithParams(uiWindow *parent, uiFileDialogParams *params)
+{
+	// TODO
+}
+
+char *uiSaveFileWithParams(uiWindow *parent, uiFileDialogParams *params)
+{
+	// TODO
 }
 
 // I would use a completion handler for NSAlert as well, but alas NSAlert's are 10.9 and higher only
--- a/darwin/stddialogs.m
+++ b/darwin/stddialogs.m
@@ -80,11 +80,15 @@ char *uiOpenFileWithParams(uiWindow *parent, uiFileDialogParams *params)
 char *uiOpenFolderWithParams(uiWindow *parent, uiFileDialogParams *params)
 {
 	// TODO
+	uiprivImplBug("Not yet implemented.");
+	return NULL;
 }
 
 char *uiSaveFileWithParams(uiWindow *parent, uiFileDialogParams *params)
 {
 	// TODO
+	uiprivImplBug("Not yet implemented.");
+	return NULL;
 }
 
 // I would use a completion handler for NSAlert as well, but alas NSAlert's are 10.9 and higher only
--- a/darwin/stddialogs.m
+++ b/darwin/stddialogs.m
@@ -32,7 +32,7 @@ static char *runSavePanel(NSWindow *parent, NSSavePanel *s)
 	return filename;
 }
 
-char *uiOpenFile(uiWindow *parent)
+char *uiOpenFile(uiWindow *parent, const char *defaultPath, const char *filter)
 {
 	NSOpenPanel *o;
 
@@ -46,7 +46,7 @@ char *uiOpenFile(uiWindow *parent)
 	return runSavePanel(windowWindow(parent), o);
 }
 
-char *uiOpenFolder(uiWindow *parent)
+char *uiOpenFolder(uiWindow *parent, const char *defaultPath)
 {
 	NSOpenPanel *o;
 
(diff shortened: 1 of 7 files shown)


cody271 (commented):
While we are now open to breaking changes in the API, only where it is truly necessary and worth the cost.. These should be added as new functions instead of breaking the API.

Also this introduces a custom (and undocumented!) string format with two slightly different parser implementations. This is a security risk against untrusted input, and not necessary at all. Should probably be an "argv" style string array. Maybe something like:
```c
_UI_EXTERN char *uiOpenFileWithFilter(uiWindow *parent, const char *defaultPath, const char **filter);
```

desttinghim:
Should the interface change to using a parameters structure, similar to `uiInitOptions` or `uiTableParams`? That would make possible changes in the future easier.

windows/stddialogs.cpp
```diff
@@ -33,6 +34,44 @@ char *commonItemDialog(HWND parent, REFCLSID clsid, REFIID iid, FILEOPENDIALOGOP
...
+
+
+	if (params != NULL) {
+		COMDLG_FILTERSPEC filterspec[8];
+		wchar_t* filter_strings[8][2];
+		int s = 0;
+		int i = 0;
+
+		// Loop over filters and convert to UTF16
+		for (s = 0; s < 8; s++) {
```
cody271, line 46: Where does the limit of 8 items come from?
(Don't see anything about an arbitrary limit in the official Windows [docs](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setfiletypes).)

unix/stddialogs.c
```diff
@@ -6,23 +6,41 @@
...
+	int s;
+
 	fcd = gtk_file_chooser_dialog_new(NULL, parent, mode,
 		"_Cancel", GTK_RESPONSE_CANCEL,
 		confirm, GTK_RESPONSE_ACCEPT,
 		NULL);
 	fc = GTK_FILE_CHOOSER(fcd);
+
+	if (params != NULL) {
+		for (s = 0; s < 8 && params->filterNames[s] && params->filterExtensions[s]; s++) {
```
cody271, line 26: I see `Must be the same length as filterNames.` in the doc string. The iteration itself looks safe, but no validation checks in the code to throw an error.

cody271:
> Should the interface change to using a parameters structure, similar to `uiInitOptions` or `uiTableParams`? That would make possible changes in the future easier.

Yes `uiFileDialogParams` looks like the way to go.

unix/stddialogs.c
desttinghim: I wasn't sure how to do error handling properly, so I added a TODO. What's the proper way to return an error here?

windows/stddialogs.cpp
desttinghim: Limit of 8 comes from the original PR, since it uses a fixed size array to pass the filter patterns and names to SetFileTypes. We should probably just allocate and free here.

unix/stddialogs.c
cody271: Use `uiprivUserBug()` for now.

darwin/stddialogs.m
```diff
@@ -70,6 +70,21 @@ static void setupSavePanel(NSSavePanel *s)
 	return runSavePanel(windowWindow(parent), s);
 }
 
+char *uiOpenFileWithParams(uiWindow *parent, uiFileDialogParams *params)
+{
+	// TODO
```
cody271, line 75:
```suggestion
	// TODO
	uiprivImplBug("Not yet implemented.");
	return NULL;
```

examples/controlgallery/main.c
```diff
@@ -203,6 +204,64 @@ static void onSaveFileClicked(uiButton *b, void *data)
...
 
+static void onOpenFileWithParamsClicked(uiButton *b, void *data)
+{
+	uiEntry *entry = uiEntry(data);
+	char *filename;
+	uiFileDialogParams fparams;
+
+	fparams.filterCount = 2;
+	fparams.filters = (uiFileDialogParamsFilter[]){
+	{"DS ROM (*.nds)", 1, (const char *[]){"*.nds"}},
```
cody271, line 215:
```suggestion
		{"DS ROM (*.nds)", 1, (const char *[]){"*.nds"}},
```

examples/controlgallery/main.c
```diff
@@ -203,6 +204,64 @@ static void onSaveFileClicked(uiButton *b, void *data)
...
+	{"Any File", 1, (const char *[]){"*.*"}},
```
cody271, line 216:
```suggestion
		{"Any File", 1, (const char *[]){"*.*"}},
```

examples/controlgallery/main.c
cody271: Also `controlgallery` should stay totally generic, not make any reference to MelonDS emulator project.

examples/controlgallery/main.c
```diff
@@ -203,6 +204,64 @@ static void onSaveFileClicked(uiButton *b, void *data)
...
+{
+	uiEntry *entry = uiEntry(data);
+	char *filename;
+	uiFileDialogParams fparams;
+
+	fparams.defaultPath = NULL;
+	fparams.defaultName = "untitled";
+	fparams.filterCount = 1;
+	fparams.filters = (uiFileDialogParamsFilter[]){
+	{"melonDS savestate (*.mln)", 1, (const char *[]){"*.mln"}},
```
cody271, line 252:
```suggestion
		{"melonDS savestate (*.mln)", 1, (const char *[]){"*.mln"}},
```

test/page5.c
```diff
@@ -101,6 +155,9 @@ uiBox *makePage5(uiWindow *pw)
 	D("Open File", openFile);
 	D("Open Folder", openFolder);
 	D("Save File", saveFile);
+	D("Open File with Params", openFileWithParams);
```
cody271, line 158:
```suggestion
	D("Open File (Advanced)", openFileWithParams);
```

ui.h
```diff
@@ -1908,6 +1908,54 @@ _UI_EXTERN void uiMenuAppendSeparator(uiMenu *m);
...
 _UI_EXTERN uiMenu *uiNewMenu(const char *name);
 
+/**
+ * File Filter struct to be passed to uiOpenFileWithParams(), uiOpenFolderWithParams(), and uiSaveFileWithParams().
+ *
+ * @struct uiFileDialogParams
+ * @ingroup dataEntry dialogWindow
+ */
+typedef struct uiFileDialogParamsFilter uiFileDialogParamsFilter;
+struct uiFileDialogParamsFilter {
```
cody271, line 1918: This data structure looks safer and more bindings friendly. 👍

windows/stddialogs.cpp
```diff
@@ -33,6 +34,62 @@ char *commonItemDialog(HWND parent, REFCLSID clsid, REFIID iid, FILEOPENDIALOGOP
...
+				filterpatterns[s] = strf(L"%s;%s", filterpatterns[s], filter.patterns[pattern]);
+				uiprivFree(old);
+			}
+
+			filternames[s] = toUTF16(filter.name);
+			filterspec[s].pszName = filternames[s];
+			filterspec[s].pszSpec = filterpatterns[s];
+		}
+
+		// TODO: assert that filterNames/filterExtensions end at the same time
```
cody271, line 72: `uiprivUserBug()`

cody271 (commented):
Nits: Still see some references to MelonDS emulator. Also missing a few more NYI errors, and the struct formatting issues.

examples/controlgallery/main.c
```diff
@@ -203,6 +204,64 @@ static void onSaveFileClicked(uiButton *b, void *data)
...
+	uiEntrySetText(entry, filename);
+	uiFreeText(filename);
+}
+
+static void onOpenFolderWithParamsClicked(uiButton *b, void *data)
+{
+	uiEntry *entry = uiEntry(data);
+	char *filename;
+
+	filename = uiOpenFolderWithParams(mainwin, NULL);
```
cody271, line 233: Why no `uiFileDialogParams` here?

desttinghim: It wasn't present in the original PR, mostly. I think the best param to test here is `defaultPath`. I think that's going to require some conditional compilation in the example to deal with windows.

matyalatte:
Hi, I tried this branch and found out the windows implementation won't work properly.
We need to fix `windows/stddialogs.cpp`.

PR #211, opened 2023-06-23
some modifications about Windows platform

Fsu0413:
1. UB when using LTO - same name in structure in different file breaks it
2. libui-ng provided ABI bug workaround does not compile using [llvm-mingw](https://github.com/mstorsjo/llvm-mingw). Since the ABI bug workaround is already provided by newer MinGW-w64 under macro let's just judge if the macro exists.


cody271:
Hi @Fsu0413, looks good! We don't actually have Clang on Windows in the CI build.. I'll have to do some more thorough testing locally before merge.

msink:
#83

PR #214, opened 2023-07-27
Fix indentation in build.yml

kojix2:
This only corrects the indentation of the YAML file.

Background:
For Ruby users who want to use the latest libui-ng, I have slightly rewritten build.yml to build the shared library for Windows and arm-linux with GitHub Actions. I noticed this indentation error when I looked at the difference between the original build.yml and the modified [pre-build.yml](https://github.com/kojix2/libui-ng/blob/pre-build/.github/workflows/build.yml).


cody271 (approved):
LGTM

PR #215, opened 2023-08-02
Add workflow_dispatch to build.yml

kojix2:
GitHub Actions will automatically remove the artifact after 90 days.
Thus the latest builds are often unavailable.

This pull request simply adds workflow_dispatch: anyone who forks the libui-ng repository will be able to run the build workflow on GitHub with the push of a button from their browser. Thus, if GitHub removes the artifact, you can immediately do the build and have a link to it ready. (However, nightly.link may not work well)

By the way, what I really want is not static libraries, but a distribution of shared libraries needed to create bindings for other languages. Let's talk about that some other time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  workflow_dispatch:
 
 jobs:
   meson-linux:


cody271:
> GitHub Actions will automatically remove the artifact after 90 days. Thus the latest builds are often unavailable.
>

Oh I hadn't noticed that the build artifacts timed out like that, good catch!
 
> This pull request simply adds workflow_dispatch: anyone who forks the libui-ng repository will be able to run the build workflow on GitHub with the push of a button from their browser. Thus, if GitHub removes the artifact, you can immediately do the build and have a link to it ready. (However, nightly.link may not work well)
>

How is `workflow_dispatch` different than just manually triggering the CI jobs from the browser already?

image

kojix2:
Thanks. I think the problem is that sometimes the [Re-run jobs] button does not appear.

When I created this pull request, Re-run button did not appear on my forked libui-ng repository. However, a new commit (#214 ) was merged and workflow was recently executed, so the button now appears.

I checked some of my repositories. In repositories where the last commit was 2-3 months ago, the Re-run jobs button tends not to show up when I open the actions tab and look at it. 
Perhaps this button disappears after a certain amount of time has passed since the last run.

cody271:
> GitHub Actions will automatically remove the artifact after 90 days.

> In repositories where the last commit was 2-3 months ago, the Re-run jobs button tends not to show up when I open the actions tab and look at it.
> Perhaps this button disappears after a certain amount of time has passed since the last run.

Yeah appears as though they expire together at the same time, so `workflow_dispatch` probably is needed.

Not finding much documentation from GitHub besides the [blog post](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch) where the feature was introduced.. what are the security implications of enabling this?

matyalatte:
Hi, I want `workflow_dispatch` to test my branch before sending PR!

> what are the security implications of enabling this?

The manual trigger is only available for owners (and collaborators, maybe) even if you set `Allow all workflows` to the actions permissions.
So, no one can mess up your repo.
We need to fork the repo if we want to run the jobs.
I mean, there is no security concerns, I think.

cody271:
> * [Securing workflow_dispatch manual workflow?](https://github.com/orgs/community/discussions/26008)

This thread matches my concerns exactly here, good find.

Okay I think its safe to enable this as is for now. We can easily add a whitelist of active contributors later, if it becomes necessary.. 

Real example given from GH discussion:
```yaml
jobs:
  PR-Comment:
    if: github.actor != 'kingthorin' && github.actor != 'wstgbot' && github.actor != 'ThunderSon' && github.actor != 'rejahrehim' && github.actor != 'victoriadrake'  
    runs-on: ubuntu-latest
    steps:
    - name: PR Comment
      uses: actions/github-script@v2
      with:
```

cody271 (approved):
LGTM

kojix2:
Using the cron feature within GitHub Actions to regularly build libui-ng might be more suitable than using workflow dispatch. I had overlooked the fact that GitHub Actions can schedule workflows to run periodically.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

PR #220, opened 2023-09-10
Tweaks to facilitate bindings from other languages

cdosoftei:
Hello, thanks for continuing the development of libui!

This pull requests adds a couple of small modifications to facilitate implementing bindings from other languages, in this case the [Ardillo extension](https://github.com/ardillo-php/ext) for PHP.

The first modification concerns destroying of UI controls, the issue being some controls could be destroyed outside of the purview of the embedding application, possibly leading to memory safety issues. In order to keep the implementing application in the loop, it implements an `onFree` callback for `uiControl` structures (and it's corresponding context as `onFreeData`).

The second modification affects structures passed as parameters to various handlers which do not carry any context data (i.e. to match them back to their corresponding structures from the embedding environment). We found that `uiArea` and `uiTableModel` structures do not provide a straightforward correlation mechanism so we added `userData` structure members and corresponding `...GetUserData` and `...SetUserData` functions to retrieve/set them.

I'm happy to maintain a synchronized fork with these modifications yet these modifications might be useful for other similar projects, hence they're presented here for your consideration.

### DIFF
--- a/common/control.c
+++ b/common/control.c
@@ -74,6 +74,11 @@ void uiFreeControl(uiControl *c)
 {
 	if (uiControlParent(c) != NULL)
 		uiprivUserBug("You cannot destroy a uiControl while it still has a parent. (control: %p)", c);
+
+	if (c->onFreeData) {
+		(*(c->onFree))(c, c->onFreeData);
+	}
+
 	uiprivFree(c);
 }
 
@@ -98,4 +103,10 @@ int uiControlEnabledToUser(uiControl *c)
 		c = uiControlParent(c);
 	}
 	return 1;
+}
+
+void uiControlOnFree(uiControl *w, void (*f)(uiControl *, void *senderData), void *data)
+{
+	w->onFree = f;
+	w->onFreeData = data;
 }
--- a/darwin/area.m
+++ b/darwin/area.m
@@ -33,6 +33,8 @@ struct uiArea {
 	uiAreaHandler *ah;
 	BOOL scrolling;
 	NSEvent *dragevent;
+
+	void *userData;
 };
 
 @implementation areaView
@@ -493,4 +495,14 @@ uiArea *uiNewScrollingArea(uiAreaHandler *ah, int width, int height)
 	a->view = a->sv;
 
 	return a;
+}
+
+void* uiAreaGetUserData(uiArea *a)
+{
+	return a->userData;
+}
+
+void uiAreaSetUserData(uiArea *a, void *userData)
+{
+	a->userData = userData;
 }
--- a/darwin/table.h
+++ b/darwin/table.h
@@ -8,6 +8,7 @@ struct uiTableModel {
 	uiTableModelHandler *mh;
 	uiprivTableModel *m;
 	NSMutableArray *tables;
+	void *userData;
 };
 struct uiTable {
 	uiDarwinControl c;
--- a/darwin/table.m
+++ b/darwin/table.m
@@ -456,3 +456,13 @@ void uiTableColumnSetWidth(uiTable *t, int column, int width)
 	else
 		[tc setWidth: width];
 }
+
+void* uiTableModelGetUserData(uiTableModel *m)
+{
+	return m->userData;
+}
+
+void uiTableModelSetUserData(uiTableModel *m, void *userData)
+{
+	m->userData = userData;
+}
--- a/ui.h
+++ b/ui.h
@@ -2062,7 +2062,7 @@ _UI_EXTERN uiArea *uiNewScrollingArea(uiAreaHandler *ah, int width, int height);
  * @returns pointer to the user data structure.
  * @memberof uiArea
  */
-void* uiAreaGetUserData(uiArea *m);
+_UI_EXTERN void* uiAreaGetUserData(uiArea *m);
 
 /**
  * Sets the user data associated with the uiTableModel.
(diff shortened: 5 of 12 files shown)


cody271:
> in this case the [Ardillo extension](https://github.com/ardillo-php/ext) for PHP.
> 

Feel free to add this to the language bindings [list](https://github.com/libui-ng/libui-ng#language-bindings).

> The first modification concerns destroying of UI controls, the issue being some controls could be destroyed outside of the purview of the embedding application, possibly leading to memory safety issues. In order to keep the implementing application in the loop, it implements an `onFree` callback for `uiControl` structures (and it's corresponding context as `onFreeData`).
> 
This is a good addition. Although anything involving memory management like this would require proper testing before merge, and I don't see any [tests](https://github.com/libui-ng/libui-ng/tree/master/test) added in this PR.

> The second modification affects structures passed as parameters to various handlers which do not carry any context data (i.e. to match them back to their corresponding structures from the embedding environment). We found that `uiArea` and `uiTableModel` structures do not provide a straightforward correlation mechanism so we added `userData` structure members and corresponding `...GetUserData` and `...SetUserData` functions to retrieve/set them.
> 
So far this hasn't been necessary in other bindings, as `uiAreaHandler`/`uiTableModelHandler` references are already user defined types. In your case the code would look something like this:

```c
zend_object_handlers AreaHandler_object_handlers;

struct zAreaHandler {
    uiAreaHandler handler;
    void *userData;
};

typedef struct zAreaHandler zAreaHandler;

zend_object* ardillo_create_AreaHandler_object(zend_class_entry *ce)
{
    ardillo_ui_AreaHandler_t *ardillo_s = ecalloc(1, sizeof(ardillo_ui_AreaHandler_t) + zend_object_properties_size(ce));
    ardillo_s->uis = (zAreaHandler *)ecalloc(1, sizeof(zAreaHandler));
    ardillo_s->uis->handler.DragBroken = ardillo_call_AreaHandler_dragBrokenHandler;
    ardillo_s->uis->handler.Draw = ardillo_call_AreaHandler_drawHandler;
    ardillo_s->uis->handler.KeyEvent = ardillo_call_AreaHandler_keyEventHandler;
    ardillo_s->uis->handler.MouseCrossed = ardillo_call_AreaHandler_mouseCrossedHandler;
    ardillo_s->uis->handler.MouseEvent = ardillo_call_AreaHandler_mouseEventHandler;
    ardillo_s->uis->userData = ardillo_s;

...
```
But I'm not familiar with PHP native extension APIs, perhaps there is a reason this approach doesn't work here?

> I'm happy to maintain a synchronized fork with these modifications yet these modifications might be useful for other similar projects, hence they're presented here for your consideration.

Please try to upstream any improvements you may have made, PRs welcome!

cdosoftei:
This is excellent feedback, thanks @cody271!

1. Just added a [new PR](https://github.com/libui-ng/libui-ng/pull/223) for the language bindings list;
2. I'll go ahead and create some relevant tests for the `onFree` control addition;
3. Regarding the `userData` back reference, I definitely overthought that `uiAreaHandler` and `uiTableModelHandler` use cases ... in fact, since they're being allocated by the implementing application, I can align them at the beginning of the embedding struct so no back reference is required.
4. I'm still unsure how to deal with `uiArea` and `uiTableModel` references; since they are allocated by libui itself, the implementer doesn't have the option to align those structs in an obvious fashion.

I've studied some of other language binding implementations and it appears there are two options to my conundrum:

(a) the embedding environment maintains some sort of a hash using the libui struct pointer as the key and the language specific structure;
(b) the `uiArea` vs `uiAreaHandler` and `uiTableModel` vs `uiTableModelHandler` abstraction is removed completely from the higher level language, i.e. behind the scenes the two structs are paired in a 1-to-1 relationship.

I've previously attempted (a) and the incurred performance penalty drove me to the implementation I've suggested in this PR.

I'm a big fan of decoupling and delegation between various entities, therefore I really like how the said objects are delimited from their respective handlers; that said, now I'm questioning if I need to carry over this abstraction in the higher level language. Based on what I've seen in other language binding implementations, it appears that other developers favoured this approach.

cody271:
> I'll go ahead and create some relevant tests for the onFree control addition;
>
👍

cody271:
> (a) the embedding environment maintains some sort of a hash using the libui struct pointer as the key and the language specific structure;
>
This has been the recommended approach for other language binding implementations so far.

> I've previously attempted (a) and the incurred performance penalty drove me to the implementation I've suggested in this PR.
>
Is this a graphics intense application?

Performance overhead of a single hash key lookup per control call or event should be negligible, most of the time.
The number of calls/events per second for typical `uiControl` usage patterns is quite low.

Complex rendering inside `uiArea` is an example of where this might be a problem, and maybe with a very large `uiTable` as well.

cdosoftei:
Thanks for the clarification.

I have indeed encountered performance issues for tables (not sure if a few hundred rows qualify as a very large table). However, I think I need to look at this again as my prior implementation used PHP's own hash table structure (it made sense for discovery purposes, but it's not really what they are designed for); this factor alone could be the reason for my prior experience.

Will report my findings soon, thanks.

cdosoftei:
Hello, it's been a while yet I finally found some time to write some tests for the first matter discussed in this request (the `onFree` callback for `uiControl` instances). I went ahead and create a PR of its own [here](https://github.com/libui-ng/libui-ng/pull/250).

I haven't wrapped my head around the second matter (addressing `uiArea` and `uiTableModel` and their respective handlers) but that's not a concern just now, so I'll go ahead and close this PR.

Thanks for your prior guidance on both matters.

PR #222, opened 2023-09-12
update documentation to provide more information about freeing functions

teras:
Functions affected: uiFreeText, uiControlDestroy, uiFreeControl

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,31 +22,31 @@ jobs:
     runs-on: ${{ matrix.os }}
     name: Ubuntu-${{ matrix.arch }}-${{ matrix.libtype }}
     steps:
-    - name: Install Meson and Linux Deps
-      run: |
-        python3 -m pip install --upgrade pip setuptools wheel
-        pip3 install meson ninja
-        sudo apt-get update -y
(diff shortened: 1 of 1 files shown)


cody271 (commented):
The header comments need to be in Doxygen format.

teras:
Fixed. Sorry for the error.

cody271:
Instead of copying my off hand comments from the issue into the docs verbatim, try to match the existing voice and style of the doc text a bit more. Keep it concise. For example, no need to prepend `this function is` to any entry.

I would also recommend [building the docs](https://github.com/libui-ng/libui-ng/blob/master/doc/doxygen.md) from source locally to test your PR.

teras:
Updated

cody271:
> Updated

Much better! 👍

cody271:
Only small nit remaining, two trailing whitespace errors in the commit:
```shell
222.patch:24: trailing whitespace.
 * @note Every time a string is returned from the library, this method should be called. 
222.patch:38: trailing whitespace.
 * 
warning: 2 lines add whitespace errors.
```
image

cody271:
Replaced with #227

PR #224, opened 2023-09-16
add placeholder for uiEntry and uiEditableCombobox

matyalatte:
Added functions
- uiEntryPlaceholder, uiEntrySetPlaceholder
- uiEditableComboboxPlaceholder, uiEditableComboboxSetPlaceholder
- uiWindowsEntryPlaceholder, uiWindowsSetEntryPlaceholder
- uiWindowsComboboxPlaceholder, uiWindowsSetComboboxPlaceholder

### DIFF
--- a/ui.h
+++ b/ui.h
@@ -813,6 +813,29 @@ _UI_EXTERN int uiEntryReadOnly(uiEntry *e);
 _UI_EXTERN void uiEntrySetReadOnly(uiEntry *e, int readonly);
 
 /**
+ * Returns the entry's placeholder.
+ *
+ * @param e uiEntry instance.
+ * @returns The placeholder text of the entry.\n
+ *          A `NUL` terminated UTF-8 string.\n
+ *          Caller is responsible for freeing the data with `uiFreeText()`.
+ * @memberof uiEntry
+ */
+_UI_EXTERN char *uiEntryPlaceholder(uiEntry *e);
+
+/**
+ * Sets text to be displayed in the entry when it is empty.
+ *
+ * @param e uiEntry instance.
+ * @param text Placeholder text.\n
+ *             A valid, `NUL` terminated UTF-8 string.\n
+ *             Data is copied internally. Ownership is not transferred.
+ * @warning Read-only entries can't not display the placeholder on Windows
+ * @memberof uiEntry
+ */
+_UI_EXTERN void uiEntrySetPlaceholder(uiEntry *e, const char *text);
+
+/**
  * Creates a new entry.
  *
  * @returns A new uiEntry instance.
@@ -1484,6 +1507,28 @@ _UI_EXTERN void uiEditableComboboxSetText(uiEditableCombobox *c, const char *tex
  */
 _UI_EXTERN void uiEditableComboboxOnChanged(uiEditableCombobox *c,
 	void (*f)(uiEditableCombobox *sender, void *senderData), void *data);
+
+/**
+ * Returns the editable combo box's placeholder.
+ *
+ * @param c uiEditableCombobox instance.
+ * @returns The placeholder text of the combo box.\n
+ *          A `NUL` terminated UTF-8 string.\n
+ *          Caller is responsible for freeing the data with `uiFreeText()`.
+ * @memberof uiEditableCombobox
+ */
+_UI_EXTERN char *uiEditableComboboxPlaceholder(uiEditableCombobox *c);
+
+/**
+ * Sets text to be displayed in the editable combo box when it is empty.
+ *
+ * @param c uiEditableCombobox instance.
+ * @param text Placeholder text.\n
+ *             A valid, `NUL` terminated UTF-8 string.\n
+ *             Data is copied internally. Ownership is not transferred.
+ * @memberof uiEditableCombobox
+ */
+_UI_EXTERN void uiEditableComboboxSetPlaceholder(uiEditableCombobox *c, const char *text);
 
 /**
  * Creates a new editable combo box.
--- a/ui_windows.h
+++ b/ui_windows.h
@@ -215,6 +215,14 @@ _UI_EXTERN int uiWindowsWindowTextWidth(HWND hwnd);
 _UI_EXTERN int uiWindowsWindowTextHeight(HWND hwnd);
 
 // TODO document
+_UI_EXTERN char *uiWindowsEntryPlaceholder(HWND hwnd, int len);
+_UI_EXTERN int uiWindowsSetEntryPlaceholder(HWND hwnd, const char *text);
+
+// TODO document
+_UI_EXTERN char *uiWindowsComboboxPlaceholder(HWND hwnd, int len);
+_UI_EXTERN int uiWindowsSetComboboxPlaceholder(HWND hwnd, const char *text);
+
+// TODO document
 // TODO point out this should only be used in a resize cycle
 _UI_EXTERN void uiWindowsEnsureMoveWindowDuringResize(HWND hwnd, int x, int y, int width, int height);
 
--- a/darwin/editablecombo.m
+++ b/darwin/editablecombo.m
@@ -134,6 +134,16 @@ void uiEditableComboboxOnChanged(uiEditableCombobox *c, void (*f)(uiEditableComb
 	c->onChangedData = data;
 }
 
+char *uiEditableComboboxPlaceholder(uiEditableCombobox *c)
+{
+	return uiDarwinNSStringToText([c->cb.cell placeholderString]);
+}
+
+void uiEditableComboboxSetPlaceholder(uiEditableCombobox *c, const char *text)
+{
+	[c->cb.cell setPlaceholderString:uiprivToNSString(text)];
+}
+
 static void defaultOnChanged(uiEditableCombobox *c, void *data)
 {
 	// do nothing
--- a/darwin/entry.m
+++ b/darwin/entry.m
@@ -150,6 +150,16 @@ void uiEntrySetReadOnly(uiEntry *e, int readonly)
 	[e->textfield setEditable:editable];
 }
 
+char *uiEntryPlaceholder(uiEntry *e)
+{
+	return uiDarwinNSStringToText([e->textfield.cell placeholderString]);
+}
+
+void uiEntrySetPlaceholder(uiEntry *e, const char *text)
+{
+	[e->textfield.cell setPlaceholderString:uiprivToNSString(text)];
+}
+
 static void defaultOnChanged(uiEntry *e, void *data)
 {
 	// do nothing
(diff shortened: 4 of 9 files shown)


matyalatte:
Here is a custom version of the controlgallery.
[controlgallery.zip](https://github.com/libui-ng/libui-ng/files/12639670/controlgallery.zip)

You can see the placeholders on the example app.
And you can see the placeholder's text on the cosole.
entries
editablecombo
read-only
stdout

cody271:
> Here is a custom version of the controlgallery.
> [controlgallery.zip](https://github.com/libui-ng/libui-ng/files/12639670/controlgallery.zip)

Please try to keep all contributions inside version control. Passing around source files like that is not recommended!

cody271:
We should add placeholders to `controlgallery`, but keep it simple:
```patch
diff --git a/examples/controlgallery/main.c b/examples/controlgallery/main.c
index 6415e4ff..e6cc9a48 100644
--- a/examples/controlgallery/main.c
+++ b/examples/controlgallery/main.c
@@ -54,14 +54,18 @@ static uiControl *makeBasicControlsPage(void)
 	uiFormSetPadded(entryForm, 1);
 	uiGroupSetChild(group, uiControl(entryForm));
 
+	uiEntry* entry = uiNewEntry();
+	uiEntrySetPlaceholder(entry, "Type here");
 	uiFormAppend(entryForm,
 		"Entry",
-		uiControl(uiNewEntry()),
+		uiControl(entry),
 		0);
+
 	uiFormAppend(entryForm,
 		"Password Entry",
 		uiControl(uiNewPasswordEntry()),
 		0);
+
 	uiFormAppend(entryForm,
 		"Search Entry",
 		uiControl(uiNewSearchEntry()),
@@ -147,6 +151,7 @@ static uiControl *makeNumbersPage()
 	uiEditableComboboxAppend(ecbox, "Editable Item 1");
 	uiEditableComboboxAppend(ecbox, "Editable Item 2");
 	uiEditableComboboxAppend(ecbox, "Editable Item 3");
+	uiEditableComboboxSetPlaceholder(ecbox, "Select an item");
 	uiBoxAppend(vbox, uiControl(ecbox), 0);
 
 	rb = uiNewRadioButtons();
@@ -264,6 +269,7 @@ static uiControl *makeDataChoosersPage(void)
 
 	button = uiNewButton("  Open File  ");
 	entry = uiNewEntry();
+	uiEntrySetPlaceholder(entry, "Select a file");
 	uiEntrySetReadOnly(entry, 1);
 	uiButtonOnClicked(button, onOpenFileClicked, entry);
 	uiGridAppend(grid, uiControl(button),
```

cody271:
> stdout
>
These sorts of checks should be added to the [test suite](https://github.com/libui-ng/libui-ng/blob/master/test/unit/entry.c), instead of the examples.

matyalatte:
> Please try to keep all contributions inside version control.

It's just for testing setter and getter methods. so I thought it shouldn't be added to the PR.

> We should add placeholders to controlgallery, but keep it simple:

ok. I'll add another commit for it.

ui_windows.h
```diff
@@ -214,6 +214,14 @@ _UI_EXTERN void uiWindowsSetWindowText(HWND hwnd, const char *text);
 _UI_EXTERN int uiWindowsWindowTextWidth(HWND hwnd);
 _UI_EXTERN int uiWindowsWindowTextHeight(HWND hwnd);
 
+// TODO document
+_UI_EXTERN char *uiWindowsEntryPlaceholder(HWND hwnd, int len);
```
cody271, line 218: I think these can be `uipriv`, and no need to expose them in the  header. For internal use only.

ui.h
```diff
@@ -812,6 +812,29 @@ _UI_EXTERN int uiEntryReadOnly(uiEntry *e);
...
+_UI_EXTERN char *uiEntryPlaceholder(uiEntry *e);
+
+/**
+ * Sets text to be displayed in the entry when it is empty.
+ *
+ * @param e uiEntry instance.
+ * @param text Placeholder text.\n
+ *             A valid, `NUL` terminated UTF-8 string.\n
+ *             Data is copied internally. Ownership is not transferred.
+ * @warning Read-only entries can't not display the placeholder on Windows
```
cody271, line 833: Hopefully this limitation can be fixed, I will try the Windows implementation next.

ui_windows.h
matyalatte: fixed at a20f26b04be1f17c7189e892a9a18f44ddb49499

ui.h
cody271: I have confirmed that it is a limitation of Win32 API that `EM_SETCUEBANNER` doesn't support read only entries or multi-line entries!

cody271:
Might as well add multi-line entry support now?
As the Unix/Darwin part is simple to implement. Underlying Windows implementation here is broken anyway.

Make a new issue specific to Windows (to keep track of it), then we can move on for now. It can be fixed in a later PR..

cody271 (commented):
Nit: `functions private`, not `fucntions private`.

ui.h
szanni, line 833:
```suggestion
 * @warning Read only entries do not display the placeholder text on Windows.
```
Spelling and consistent punctuation.

And I confirmed as well that `EM_SETCUEBANNER` does not play nice with `ES_READONLY`.

test/unit/entry.c
```diff
@@ -123,6 +140,7 @@ int entryRunUnitTests(void)
 		entryUnitTests(entryTextDefault),
 		entryUnitTests(entrySetText),
 		entryUnitTests(entrySetTextNoCallback),
+		entryUnitTests(entrySetPlaceholder),
 		entryUnitTests(entryReadOnlyDefault),
 		entryUnitTests(entrySetReadOnly),
```
szanni, line 145:
```suggestion
		entryUnitTests(entrySetReadOnly),
		entryUnitTests(entryReadOnlySetPlaceholder),
		entryUnitTests(entryPlaceholderSetReadOnly),
```

szanni: And add the following two tests to line 87:
```c
// win32: Make sure the placeholder gets set in read only mode, even though not visually displayed.
static void entryReadOnlySetPlaceholder(void **state)
{
	uiEntry **e = uiEntryPtrFromState(state);
	const char *text = "Text";
	char *rv;

	uiEntrySetReadOnly(*e, 1);
	uiEntrySetPlaceholder(*e, text);
	assert_int_equal(uiEntryReadOnly(*e), 1);
	rv = uiEntryPlaceholder(*e);
	assert_string_equal(rv, text);
	uiFreeText(rv);
}

// win32: Make sure the placeholder is retained and read only mode settable with a placeholder set.
static void entryPlaceholderSetReadOnly(void **state)
{
	uiEntry **e = uiEntryPtrFromState(state);
	const char *text = "Text";
	char *rv;

	uiEntrySetPlaceholder(*e, text);

	uiEntrySetReadOnly(*e, 1);
	assert_int_equal(uiEntryReadOnly(*e), 1);
	rv = uiEntryPlaceholder(*e);
	assert_string_equal(rv, text);
	uiFreeText(rv);

	uiEntrySetReadOnly(*e, 0);
	assert_int_equal(uiEntryReadOnly(*e), 0);
	rv = uiEntryPlaceholder(*e);
	assert_string_equal(rv, text);
	uiFreeText(rv);
}
```

test/unit/entry.c
```diff
@@ -51,6 +51,23 @@ static void entrySetTextNoCallback(void **state)
 	uiEntrySetText(*e, "Text 2");
 }
 
+static void entrySetPlaceholder(void **state)
```
szanni, line 54:
```suggestion
static void entryPlaceholderDefault(void **state)
{
	uiEntry **e = uiEntryPtrFromState(state);
	const char *text = "";
	char *rv;

	rv = uiEntryPlaceholder(*e);
	assert_string_equal(rv, text);
	uiFreeText(rv);
}

static void entrySetPlaceholder(void **state)
```

szanni: This currently crashes on all platforms and needs fixing.

windows/uipriv_windows.hpp
szanni, line 1: None of the other functions are prefixed with `uipriv`. Do we wan't to specify any rules around this?

szanni (changes requested):
I have so far only checked `uiEntry` and added a few more test cases - one of which crashes *hard* on all platforms.

From an API perspective it is the question of what the default value of `uiEntryPlaceholder` should be. An empty string or `NULL`?
And do we want to support the setting of `NULL` or do we unset via an empty string?

I haven't checked but the crash on all three platforms suggests they return some type of `NULL` value initially. The test I added expects an empty string, as that is what `uiEntryText` returns by default.
I am not entirely opposed to `NULL` values either. The docs would need to be changed to reflect that though.

I would love to also have a `qa` test for the interaction of read only and placeholder text.
And I would wouldn't mind having two different PRs for `uiEntry` and `uiCombobox` as this is a lot of code - I do understand they have a fair amount of overlap though on win32.

windows/uipriv_windows.hpp
cody271: Other functions just predate the policy change. Original discussion here:
> https://github.com/andlabs/libui/issues/308
>
>> * prefix all non-public names that aren't `static` with `uipriv`

ui.h
matyalatte: > Read-only -> Read only

I can see some comments that use "read-only" in the repo. (e.g. [windows/multilineentry.cpp](https://github.com/libui-ng/libui-ng/blob/257f2664b18df67778b68318125756ae691ce1e2/windows/multilineentry.cpp#L4))
But should I use "Read only" in the document?

> can't -> do

I thought I fixed it already but I seemed to revert it somehow...

test/unit/entry.c
matyalatte: fixed at ac3ce075957f856094450010c46a541907a584ce

ui.h
matyalatte: fixed at 3302ef9d668b23cbecfa6b37b46a7c9cd0e13d56

test/unit/entry.c
matyalatte: I noticed I don't need to edit `entry.cpp` to pass these tests.
I mean, `EM_SETCUEBANNER` works internally with `ES_READONLY`.
I made another branch to test them but it passed the CI worklfow.

matyalatte:
> Might as well add multi-line entry support now?

It requires `WM_*` events to catch focus events and to change the text color.
So, I think we should work on the event handling features first.
like `uiEntryOnFocused`, `uiEntrySetTextColor`, etc.

> Make a new issue specific to Windows (to keep track of it), then we can move on for now. It can be fixed in a later PR.

Ok, I'll make another issue like "Read only entries do not display placeholders on Windows" after this PR is merged.

PR #233, opened 2023-10-07
Fix windows entry long text scroll.

### LINKED ISSUE #232: Unnecessary auto scrolling with uiEntrySetText on Windows
I want to set default vaules to `uiEntry`.
But when you use `uiEntrySetText` at launch, the app scrolls the edit control horizontally even if it has enough width to show whole text.

As an example, I set a to z as a default value in the control gallery on Windows 10.
```c
uiEntry* e = uiNewEntry();
uiEntrySetText(e, "abcdefghijklmnopqrstuvwxyz");
uiFormAppend(entryForm,
  "Entry",
  uiControl(e),
  0);
```

Then, it showed from m at launching.
unexpected

What I expected was like this.
expected

Is there any workaround for it?
I noticed you can pass the default value to `uiWindowsEnsureCreateControlHWND` to fix.
But it requires a new function like `uiNewEntryWithParams`...

szanni:
Fixes #232 

Long text entries display in a weirdly scrolled way on Windows upon first window creation. This is due to the cursor being placed at the end of the text but control dimension still being unknown.

Fix by placing the cursor only if the uiEntry has the focus as cursor placement is otherwise determined by the user upon clicking the control.

Verify: qa > uiEntry > 4. Entry Long Text

~~I would still hold off merging this until we can determine why the cursor position is being set in the first place (introduced in #102).~~
The only use case I can think of is when the `uiEntry` in question is already focused. @adamierymenko was this your original use case? Setting the text when the user is already editing the text within the control? If that is the case I will add another `qa` test.
Otherwise my cursor is always placed right under my mouse cursor once I click or at the end if the control is bigger (Windows 7 & wine 8.16).

### DIFF
--- a/test/qa/entry.c
+++ b/test/qa/entry.c
@@ -154,3 +154,24 @@ uiControl* searchEntryOnChanged()
 	return uiControl(vbox);
 }
 
+const char *entryLongTextGuide() {
+	return
+	"1.\tYou should see a text entry box containing the text\n"
+	"\t`abcdefghijklmnopqrstuvwxyz`. No scrolling on your part should be\n"
+	"\tnecessary to see the entire string.\n"
+	;
+}
+
+uiControl* entryLongText()
+{
+	uiBox *vbox;
+	uiEntry *entry;
+
+	vbox = uiNewVerticalBox();
+	entry = uiNewEntry();
+	uiEntrySetText(entry, "abcdefghijklmnopqrstuvwxyz");
+	uiBoxAppend(vbox, uiControl(entry), 0);
+
+	return uiControl(vbox);
+}
+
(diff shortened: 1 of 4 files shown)


szanni:
Upon further testing and reflection I have come to the conclusion that this is safe to merge.
I can not come up with any test scenario that breaks anything with this patch.

Is does however fix the issue #232 as can be seen in `qa > uiEntry > 4.` and `qa > uiEntry > 5`.

cody271:
> (Windows 7 & wine 8.16)
> It does however fix the issue https://github.com/libui-ng/libui-ng/issues/232 as can be seen in qa > uiEntry > 4. and qa > uiEntry > 5.

Confirmed here on Win 11, this patch fixes both `4. Entry Long Text` and `5. Entry Overflow Text`.

cody271 (approved):
LGTM

PR #234, opened 2023-10-25
doc: Clarify what the orientation means in uiSeparator.

szanni:
Small fix to the docs as I've been bitten by this now, twice. The naming is not intuitive to me but ok, so be it.
Is it separating vertical/horizontal elements or drawing a vertical/horizontal line to separate? This patch should clarify that.

Interestingly enough using the wrong uiSeparator seemingly makes no difference on unix and windows but breaks the layout completely on darwin.

### DIFF
--- a/ui.h
+++ b/ui.h
@@ -1308,7 +1308,7 @@ typedef struct uiSeparator uiSeparator;
 #define uiSeparator(this) ((uiSeparator *) (this))
 
 /**
- * Creates a new horizontal separator.
+ * Creates a new horizontal separator to separate controls being stacked vertically.
  *
  * @returns A new uiSeparator instance.
  * @memberof uiSeparator @static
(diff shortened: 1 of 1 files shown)


cody271 (approved):
LGTM

rubyFeedback:
Note:

"Is it separating vertical/horizontal elements or drawing a vertical/horizontal line to separate?"

I often get confused by this in GTK too. It is possible to use a resize handle, to resize e. g. two
widgets via horizontal, or via a vertical bar. So, if you have a horizontal bar, you can move the
mouse up and down to change the position; accordingly the two widgets will then change
their size, to fit to that bar. My brain never ever remembers the API - I always end up using
the wrong one like 50% of the time, until I decided that, rather than remembering the name
of the API, I simply aliased it (in ruby) towards "drag_mouse_via_up_and_down_movement".

This is much longer, but when I see it I know what this means. Aka this is a horizontal bar
actually, because only the horizontal bar can be dragged up and down to re-size the
two widgets.

PR #236, opened 2023-10-27
qa: Fix layout.

szanni:
Most of the tests had a strange layout up until now. See `qa > uiButton` for example, with the button either stretched (windows, unix) or weirdly in the center (darwin).
This was due to the controls being placed in a stretchy vbox. These two patches introduce another non stretchy sub-vbox and remove some code duplication while at it.

Unix before:
image

Unix after:
image


cody271:
Darwin before:

before

Darwin after:

after

cody271 (approved):
LGTM

PR #237, opened 2023-10-27
qa: Add manual test for uiSpinboxOnChanged().

szanni:
This is in preparation for the `uiDragDestination` PR that is still in the works, is massive, and that will require some refactoring for the some of the darwin components.
To make sure I don't break anything and to not blow up the `uiDragDestination` code even more, here one of the independent test cases.


cody271 (approved):
LGTM

Issue #216, opened 2023-08-18
Layout Issues


AndyObtiva:
I have discovered a few layout issues on the Mac (maybe they happen on other OS'es too).

1. If I nest multiplle area controls inside a grid layout (right under the window), when running the app, only one grid shows up
2. If I nest a table with an area inside a horizontal box or vertical box layout (right under the window), and I set stretchy to false for the table, when running the app, I only see the area taking up all space. Strangely enough, if I try to resize the window, the area disappears and the table shows up to take all available space.
3. Number 2 also happens if I have a multiline entry (or non-wrapping multiline entry) and an area inside a horizontal box or vertical box

These issues limit the ability to build highly sophisticated graphical user interfaces.

Interestingly, if I use a label or standard entry with an area in a horizontal or vertical box, I don't have issues. The layouts work as advertized.

cody271:
> I have discovered a few layout issues on the Mac (maybe they happen on other OS'es too).

Are these Mac only issues, or present on other platforms?

Please provide some actual code so we can reproduce these layout issues.

szanni:
Are you using any `uiSeparator`s in your code to visually separate your elements?
I've had massive layout breakage using the wrong one on darwin systems, seemingly somewhat dependent on what type of `uiControl` I was stackin.

See  the doc fix  in #234.

szanni:
Are you using any `uiSeparator`s in your code to visually separate your elements?
I've had massive layout breakage using the wrong one on darwin systems, seemingly somewhat dependent on what type of `uiControl` I was stackin.

See  the doc fix  in #234.

AndyObtiva:
I don’t believe I had any uiSeparator in my GUI. 

I haven’t had a chance to test my encountered issues outside of the Mac, and I didn’t have a chance to write code examples outside of Glimmer DSL for LibUI, but I figured the descriptions I provided were clear enough to reproduce with your own C code.

cody271:
Test case doesn't have to be C code. Ruby is okay too.

kojix2:
Grid layouts are still having problems with MacOS.
I asked DeepWiki to investigate this issue.

The root cause is that it is implemented in NSView and Auto Layout for backward compatibility, and the implementation is complex and does not work well.

Apple has provided NSGridView since macOS 10.12 (Sierra). If we use NSGridView, we could greatly simplify the current implementation and possibly solve the layout issues while improving performance.

[...]

PR #241, opened 2023-10-30
Add manual tests for uiRadioButtonsOnSelected() and fix darwin and windows.

### LINKED ISSUE #240: darwin+windows: uiRadioButtonsOnSelected() fires again when clicking a selected element.
`uiRadioButtonsOnSelected()`  fires again when clicking an already selected element. This is the case for both darwin (11.7 Big Sur) and windows (Windows 7).

On my unix system this does not happen, which IMO is the correct behavior. `OnSelected` is maybe not quite as clear as some other controls that have `OnChanged`, but I believe the same should apply: The callback should only triggered on an actual change.
Thoughts?

Check out [qa > uiRadioButtons](https://github.com/szanni/libui-ng/tree/qa-radiobuttons) to test. I documented the behavior I expect in within.

szanni:
Add manual tests for uiRadioButtonsOnSelected()  as we can currently not do so via unit tests.

Fix both darwin and windows that will trigger the `uiRadioButtonsOnSelected()` callback on element that are already selected. This matches the behavior of other controls within the library and is the default behavior on unix.

Fixes #240 

Looking at this test on darwin does give me the impression that something in the layout translation (autolayout to uiBox) is severely broken. Related to #216

### DIFF
--- a/darwin/radiobuttons.m
+++ b/darwin/radiobuttons.m
@@ -25,6 +25,7 @@ struct uiRadioButtons {
 	radioButtonsDelegate *delegate;
 	void (*onSelected)(uiRadioButtons *, void *);
 	void *onSelectedData;
+	int selected;
 };
 
 @implementation radioButtonsDelegate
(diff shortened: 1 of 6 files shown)


matyalatte:
Confirmed the issue on Windows 10 and macOS 10.15.

https://github.com/libui-ng/libui-ng/assets/69258547/ec685df9-1041-419c-a033-2674a731c50a

And confirmed that this PR fixed it.

https://github.com/libui-ng/libui-ng/assets/69258547/2a507a77-deee-4d48-a52d-49905a93ab9e

The qa test works fine on Ubuntus 20.04 as well.

And the code looks good to me.

cody271:
Confirmed here as well this PR fixes Darwin/Windows, Unix is unaffected.

cody271 (approved):
> And the code looks good to me.

LGTM. Appreciate the thorough review!

PR #243, opened 2023-11-02
Clean up examples by using global includes and remove dates.

szanni:
The examples should reflect external use and hence use `#include <ui.h>`.
Remove the visual clutter of the data comments while at it as well.


cody271:
> Remove the visual clutter of the date comments while at it as well.

Yes there was never any need for this, we have version control.

cody271 (approved):
LGTM

PR #249, opened 2023-11-08
ci: Install meson and ninja using apt instead of pip

kojix2:
Hi!
I noticed that recently, there is no longer a need for a workaround to install meson and ninja using pip instead of apt on  Ubuntu. (My understanding is that pip was used because the latest meson and ninja were not available in APT. Sorry if I am wrong.)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,10 +25,8 @@ jobs:
     steps:
       - name: Install Meson and Linux Deps
         run: |
-          python3 -m pip install --upgrade pip setuptools wheel
-          pip3 install meson ninja
           sudo apt-get update -y
-          sudo apt-get install -y libgtk-3-dev xvfb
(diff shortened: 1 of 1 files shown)


cody271:
> My understanding is that pip was used because the latest meson and ninja were not available in APT. Sorry if I am wrong.

The reason to pull from pip is that an LTS repo like in Ubuntu is going to be years out of date, *most* of the time.
Did you check the build logs to see the actual difference in version numbers?

kojix2:
The github workflows are a kind of code that many people refer to.
Using an older version of the OS package manager simplifies the workflow and makes it easier for everyone to build libuii-ng. As of December 2023, there has been no release of libuii-ng and probably never will be. This is somewhat unfortunate, but not surprising considering the difficulty of releasing an official version. But instead we need to make libui-ng easier to build. This was the motivation for writing this pull request. However, it seems that in the meson community it is more common to install using pip rather than apt. This issue is closed.

PR #245, opened 2023-11-02
Add uiDragDestination for text and file drop.

szanni:
Add text and file drop for all uiControls.

This introduces new types:
- uiDragDestination: Holds accepted content types and callbacks
- uiDragContext: Context related to cursor position, supported drag operations and a reference to the dragged data itself
- uiDragData: the actual dragged data

You can then register the drag destination with a uiControl:
`uiControlRegisterDragDestination()`.

The `Drag and drop` documentation included should give you a good overview on the basic workings.

I currently only test `uiLabel` as a drag destination. All other controls *should* work but are untested by me (apart from `uiWindow`).
Any suggestions on how to go about testing all controls? Do a drag destination test for each control within the `qa` tester? Most controls are not tested within yet.

I would personally like to merge the other `qa` tests first to make sure I didn't break anything while refactoring the darwin code.

I believe this should also close/supersede [old PR #378](https://github.com/andlabs/libui/pull/378) as we already have `uiWindowOnContentSizeChanged()` , `uiWindowOnFocusChanged()`, `uiWindowOnClosing()` implemented upstream and file drop support will be provided by this PR.

Fixes #217 - at least the initial request. Custom mime types are not included in this PR, but this should be possible by adding a `uiDragDestinationSetAcceptedStringTypes()` or similar without interfering with the predefined types API, that most people will probably use.
QT for example provides a [conversion utility ](https://doc.qt.io/qt-6/qutimimeconverter.html) to convert from custom mimes to built in types. To me that honestly seems somewhat brittle. My thought are:
Provide a predefined types API via `uiDragDestinationAcceptedTypes`.
Provide a mime type API via `uiDragDestinationAcceptedStringTypes` and filter out the overlapping predefined types.

Reviews welcome. This is a massive change set.

PR #250, opened 2023-11-11
onFree callback for uiControl instances

cdosoftei:
This pull request concerns the destroying of UI controls, the issue being some controls could be destroyed outside of the purview of the embedding application, possibly leading to memory safety issues. In order to keep the implementing application aware of controls' lifecycle/status, the PR implements an `onFree` callback for `uiControl` structures (and its corresponding context as `onFreeData`).

This topic was originally discussed [here](https://github.com/libui-ng/libui-ng/pull/220#issue-1889287906).

### DIFF
--- a/test/unit/main.c
+++ b/test/unit/main.c
@@ -32,10 +32,18 @@ int unitTestSetup(void **_state)
 	return 0;
 }
 
+void unitControlOnFree(uiControl *c, void *data)
+{
+	assert_non_null(c);
+	assert_non_null(data);
+	assert_ptr_equal(c, ((struct state *)data)->w);
+}
+
 int unitTestTeardown(void **_state)
 {
 	struct state *state = *_state;
 
+	uiControlOnFree(uiControl(state->w), unitControlOnFree, state);
 	uiWindowSetChild(state->w, uiControl(state->c));
 	uiControlShow(uiControl(state->w));
 	//uiMain();
--- a/common/control.c
+++ b/common/control.c
@@ -74,6 +74,11 @@ void uiFreeControl(uiControl *c)
 {
 	if (uiControlParent(c) != NULL)
 		uiprivUserBug("You cannot destroy a uiControl while it still has a parent. (control: %p)", c);
+
+	if (c->onFree) {
+		(*(c->onFree))(c, c->onFreeData);
+	}
+
 	uiprivFree(c);
 }
 
@@ -98,4 +103,10 @@ int uiControlEnabledToUser(uiControl *c)
 		c = uiControlParent(c);
 	}
 	return 1;
+}
+
+void uiControlOnFree(uiControl *w, void (*f)(uiControl *, void *senderData), void *data)
+{
+	w->onFree = f;
+	w->onFreeData = data;
 }
--- a/ui.h
+++ b/ui.h
@@ -116,6 +116,9 @@ struct uiControl {
 	int (*Enabled)(uiControl *);
 	void (*Enable)(uiControl *);
 	void (*Disable)(uiControl *);
+
+	void (*onFree)(uiControl *, void *);
+	void *onFreeData;
 };
 // TOOD add argument names to all arguments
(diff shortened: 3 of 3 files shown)


test/unit/main.c
```diff
@@ -32,10 +32,18 @@ int unitTestSetup(void **_state)
...
+	assert_non_null(c);
+	assert_non_null(data);
+	assert_ptr_equal(c, ((struct state *)data)->w);
+}
+
 int unitTestTeardown(void **_state)
 {
 	struct state *state = *_state;
 
+	uiControlOnFree(uiControl(state->w), unitControlOnFree, state);
```
cody271, line 46: This test doesn't prove that `unitControlOnFree()` is ever actually called.

cdosoftei: Indeed; I'm unsure how to address it without some sort of state tracking. Since there's already a state variable, I've added a flag to its struct definition and using it to explicitly check whether the `onFree` callback is indeed called or not.

test/unit/unit.h
```diff
@@ -31,6 +31,7 @@ int progressBarRunUnitTests(void);
 struct state {
 	uiWindow *w;
 	uiControl *c;
+	int onFreeCalled;
```
szanni, line 34:
```suggestion
```

test/unit/main.c
```diff
@@ -32,16 +32,27 @@ int unitTestSetup(void **_state)
...
+	state->onFreeCalled = 0;
 	uiWindowSetChild(state->w, uiControl(state->c));
 	uiControlShow(uiControl(state->w));
 	//uiMain();
 	uiMainSteps();
 	uiMainStep(1);
 	uiControlDestroy(uiControl(state->w));
+	assert_int_equal(state->onFreeCalled, 1);
```
szanni, line 55:
```suggestion
```

test/unit/main.c
```diff
@@ -32,16 +32,27 @@ int unitTestSetup(void **_state)
...
+	assert_ptr_equal(c, ((struct state *)data)->w);
+	((struct state *)data)->onFreeCalled = 1;
+}
+
 int unitTestTeardown(void **_state)
 {
 	struct state *state = *_state;
 
+	uiControlOnFree(uiControl(state->w), unitControlOnFree, state);
+	state->onFreeCalled = 0;
```
szanni, line 48:
```suggestion
```

test/unit/main.c
```diff
@@ -32,16 +32,27 @@ int unitTestSetup(void **_state)
 	return 0;
 }
 
+void unitControlOnFree(uiControl *c, void *data)
+{
+	assert_non_null(c);
+	assert_non_null(data);
+	assert_ptr_equal(c, ((struct state *)data)->w);
+	((struct state *)data)->onFreeCalled = 1;
```
szanni, line 40:
```suggestion
	function_called();
```

szanni (changes requested):
I still struggle to understand why the Ardillo PHP binding needs an `OnFree` callback. How has this not come up as an issue for other bindings? Do they not care and simply leak memory? Or is the need for `OnFree` rather the need stemming from design choices made within the `Ardillo` binding? Because I would be surprised if PHP is too limited in language features to do without - when compared to other bindings.

Apart from that, thanks for providing unit tests!
This is sadly however not the way to test this.
For guaranteeing the invocation of a function call  you want to use `expect_function_calls(unitControlOnFree, 1);`  and `function_called();`. I provided hints in the PR review on how to do so with cmocka.

I personally am not really a fan of intermingling this testing with the setup and tear down of each individual test. I would prefer this to be done for each control individually. There may be a discussion to be had around this in general. I am facing the same dilemma in #245 which implements drag and drop for each uiControl.

Maybe there is a way to create a `controlUnitTest` setup and teardown that can be called by each individual `uiControl` as the first unit test case? Otherwise we'll have the entire test suite break in case something around the `onFree` callback goes wrong.

PR #256, opened 2023-12-02
WIP Tooltips

petabyt:
Adapted from https://github.com/libui-ng/libui-ng/commit/b65e161f941252a1f49ce385987792ddf2268637

Implements
```
/**
 * Sets the control tooltip.
 *
 * @param c uiControl instance.
 * @param tooltip Control tooltip.\n
 *             A valid, `NULL` terminated UTF-8 string.\n
 *             Data is copied internally. Ownership is not transferred.
 * @note Setting `NULL` resets the tooltip to the default value.
 * @memberof uiControl
 */
_UI_EXTERN void uiControlSetTooltip(uiControl *c, const char *tooltip);
```
Tested only on GTK3 and WinForms. I tested only on buttons. I might write some tests to make sure it works on other widgets.

### DIFF
--- a/windows/tooltip.cpp
+++ b/windows/tooltip.cpp
@@ -1,0 +1,45 @@
+#include "uipriv_windows.hpp"
+
+static void *CreateTooltip(HWND hparent, const wchar_t* text) {
+    HWND hwndTT = CreateWindowEx(WS_EX_TOPMOST, TOOLTIPS_CLASS, NULL,
+        WS_POPUP | TTS_NOPREFIX | TTS_ALWAYSTIP,
+        CW_USEDEFAULT, CW_USEDEFAULT, CW_USEDEFAULT, CW_USEDEFAULT,
+        hparent, NULL, hInstance, NULL);
+
+    if (!hwndTT) {
+        printf("ERROR: Failed to create tooltip window.");
+        return NULL;
+    }
+
+    SetWindowPos(hwndTT, HWND_TOPMOST, 0, 0, 0, 0,
+                 SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE);
+
+    TTTOOLINFO ti = { 0 };
+    ti.cbSize = sizeof(ti);
+    ti.uFlags = TTF_SUBCLASS;
+    ti.hwnd = hparent;
+    ti.hinst = hInstance;
+    ti.lpszText = (LPWSTR)text;
+    GetClientRect(hparent, &ti.rect);
+    ti.rect.right = 400;
+
+    if (!SendMessage(hwndTT, TTM_ADDTOOL, 0, (LPARAM)&ti)) {
+        logLastError(L"Failed to set rect to tooltip window.");
+        uiWindowsEnsureDestroyWindow(hwndTT);
+        return NULL;
+    }
+    return hwndTT;
+}
+
+void uiControlSetTooltip(uiControl *c, const char *tooltip) {
+	if (tooltip == NULL) {
+		if (uiWindowsControl(c)->tooltip == NULL) return;
+		uiWindowsEnsureDestroyWindow((HWND)uiWindowsControl(c)->tooltip);
+		return;
+	}
+    wchar_t *wtext = toUTF16(tooltip);
+    void *ptr = CreateTooltip((HWND)uiControlHandle(c), wtext);
+    uiprivFree(wtext);
+
+	uiWindowsControl(c)->tooltip = ptr;
+}
--- a/darwin/meson.build
+++ b/darwin/meson.build
@@ -46,6 +46,7 @@ libui_sources += [
 	'darwin/table.m',
 	'darwin/tablecolumn.m',
 	'darwin/text.m',
+	'darwin/tooltip.m',
 	'darwin/undocumented.m',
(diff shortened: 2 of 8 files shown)


matyalatte:
Thanks for your contributions!

My windows imprementation won't work for containers (e.g. uiBox) and some controls that have children (e.g. uiSpinbox.)
And it uses `400` as tooltip width. (It's the window size of my app.)

So, we should test if they are fixed or not.

windows/tooltip.cpp
```diff
@@ -0,0 +1,45 @@
...
+                 SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE);
+
+    TTTOOLINFO ti = { 0 };
+    ti.cbSize = sizeof(ti);
+    ti.uFlags = TTF_SUBCLASS;
+    ti.hwnd = hparent;
+    ti.hinst = hInstance;
+    ti.lpszText = (LPWSTR)text;
+    GetClientRect(hparent, &ti.rect);
+    ti.rect.right = 400;
```
matyalatte, line 24: Here.
I remember my APP was not able to assign correct size to resizable controls without this line.
You should test how it works when you remove this line.

szanni:
> Tested only on GTK3 and WinForms. I tested only on buttons. I might write some tests to make sure it works on other widgets.

We definitely would need more tests, I'm wondering how to best go about this in general as I am facing a similar task with the drag and drop in #245. Suggestion welcome.

Your code will most likely fail around `uiSlider`  (and most likely other widgets) as `uiSlider` already has some tooltip support.

petabyt:
Also it would fail on uiWindow too. So maybe it would be a good idea to just silently fail (or return an error?) on invalid widgets.
Or maybe it would be a good idea to limit tooltips to a few widgets? Would make testing easier.

For tests, I would add a page in the test app with a tooltip for all widgets, test updating tooltips, removing them, and also add some usage in the example programs

rubyFeedback:
This would be pretty cool to have. Personally I do not need libui as such (I solve all my computer-related issues via self-written ruby scripts), but for elderly people who need a GUI, being able to show tooltips can be immensely helpful.

rubyFeedback:
> Or maybe it would be a good idea to limit tooltips to a few widgets? Would make testing easier.

I would just limit it to a few widgets for now. It is better to have partial support for this than no
support. The documentation can always state for which widgets it works right now.

Some widgets are more important than others. For instance, I think I would need this the
most for button widgets. And perhaps simple text-widgets too, but buttons are more
important in as much that a user may want to click on it, and may want to get some
ad-hoc info what clicking this button does when clicked, but before clicking it.

petabyt:
Buttons are definitely a must, but somebody might want them on dropdowns, radios, maybe labels? But I don't think anybody would want them on a uiWindow, uiMenu, uiBox, (and uiSlider)

szanni:
> Buttons are definitely a must, but somebody might want them on dropdowns, radios, maybe labels? But I don't think anybody would want them on a uiWindow, uiMenu, uiBox, (and uiSlider)

From experience people usually end up wanting them for all controls - for all kinds of strange reasons. uiMenu is not a uiControl and probably impossible to implement on macOS for example but I'd be surprised if nobody came along wanting a tooltip for uiBox or uiWindow.
I don't believe these need to be there for all controls from the start though. Solely a thorough check upon creation and a `uiprivUserBug()` if the control should be unsupported.

matyalatte:
@petabyt
Can you adapt [latest commits of my branch](https://github.com/matyalatte/libui-ng/commits/tuw/)?
My windows imprementation was made only for my apps, so it has many problems.

At least, these four commits should be merged.
- Converted indentation to tabs (https://github.com/matyalatte/libui-ng/commit/80873c630fb0ed66be7c4a669b490cbd2fd0e705)
- Replaced printf with logLastError (https://github.com/matyalatte/libui-ng/commit/858a3ce122c4e2f6231affe7786e96760d7ae0ee)
- Fixed a compiler error for macOS, rename a parameter in unix/tooltip.c (https://github.com/matyalatte/libui-ng/commit/183fa0c9216a0c19c576a21219d51d7df228d836)
- Fixed a bug that controls can not resize tooltip area (https://github.com/matyalatte/libui-ng/commit/b542bbbee5869411b1d53d762ef145f6a6b6d8c0)

Idk if this should be merged as it is, but you should release hwnd for tooltips.
- Release tooltip with uiControlDestroy (https://github.com/matyalatte/libui-ng/commit/499d46b7bb1e73ff1810f4e0d64a7e32ae0d4b27)

These commits are still experimental but you can support more controls like this.
- Added support for uiSpinbox (https://github.com/matyalatte/libui-ng/commit/85513f874e540e80184d3671344c54301e60678d)
- Added support for uiRadioButtons (https://github.com/matyalatte/libui-ng/commit/dabc8276614faea5d8643bcb7fe97e4d34f9e8ff)

petabyt:
> @petabyt Can you adapt [latest commits of my branch](https://github.com/matyalatte/libui-ng/commits/tuw/)?

Looks good, maybe you can make a new separate PR?

matyalatte:
@petabyt 
I made another PR #266 with the new commits.
You can use it as a base branch if you'll work on the tooltip API more.

PR #257, opened 2023-12-07
add uiImageBuffer APIs

matyalatte:
This is revival of #55.
I added support for macOS.
(But I'm novice at the core graphics. I need someone can review it.)

And I made a demo for the new APIs. You can see how they work.
https://github.com/matyalatte/libui-sprite-demo

https://github.com/matyalatte/libui-sprite-demo/assets/69258547/41914efa-f9a0-4766-a105-301dfda18dc7

As mentioned in the old PR, they are not for [bit blitting](https://en.wikipedia.org/wiki/Bit_blit), but for drawing images with `uiDrawTransform`.
like this

https://github.com/libui-ng/libui-ng/assets/69258547/f56b2ccf-9d30-473e-8f14-a8e39ce695ad


matyalatte:
I noticed the old uiImageBuffer uses BGRA.
So, I fixed the color space to RGBA.

I also added QA tests and API documentation.

QA description
image_buffer_qa_desc

Result on Windows 10
image_buffer_qa_window

Result on macOS 10.15
image_buffer_qa_darwin

Result on Ubuntu 20.04
image_buffer_qa_unix

matyalatte:
Added unit tests and fixed nits in `test/qa/image_buffer.c`.

darwin/draw.m
```diff
@@ -452,3 +452,88 @@ void uiDrawRestore(uiDrawContext *c)
...
+	}
+
+	CGImageRelease(buf->img);
+	buf->img = CGBitmapContextCreateImage(buf->buf);
+}
+
+void uiImageBufferDraw(uiDrawContext *c, uiImageBuffer *buf, uiRect *srcrect, uiRect *dstrect, int filter)
+{
+	if (srcrect->Width == 0 || srcrect->Height == 0)
+		return;  // avoid deviding by zero
```
cody271, line 516:
```suggestion
		return;  // avoid dividing by zero
```

ui.h
```diff
@@ -2206,6 +2215,58 @@ _UI_EXTERN void uiDrawClip(uiDrawContext *c, uiDrawPath *path);
...
+_UI_EXTERN void uiImageBufferUpdate(uiImageBuffer *buf, const void *data);
+
+/**
+ * Draws the image buffer on the draw context.
+ *
+ * @param c uiDrawContext instance.
+ * @param buf uiImageBuffer instance.
+ * @param scrrect The size and position in the image buffer.
+ * @param dstrect The size and position in the draw context.
+ * @param filter `TRUE` to use linear interpolation, `FALSE` to use nearest neighbor.
```
cody271, line 2257: Boolean type here is ambiguous. A new enum for interpolation modes would better match the native APIs?

unix/draw.c
```diff
@@ -141,3 +141,82 @@ void uiDrawRestore(uiDrawContext *c)
...
+		dst += buf->Stride;
+	}
+
+	cairo_surface_mark_dirty(buf->buf);
+}
+
+void uiImageBufferDraw(uiDrawContext *c, uiImageBuffer *buf, uiRect *srcrect, uiRect *dstrect, int filter)
+{
+	if (srcrect->Width == 0 || srcrect->Height == 0)
+		return;  // avoid deviding by zero
```
cody271, line 198:
```suggestion
		return;  // avoid dividing by zero
```

unix/window.c
```diff
@@ -383,6 +383,7 @@ uiWindow *uiNewWindow(const char *title, int width, int height, int hasMenubar)
 	gtk_widget_set_halign(w->childHolderWidget, GTK_ALIGN_FILL);
 	gtk_widget_set_vexpand(w->childHolderWidget, TRUE);
 	gtk_widget_set_valign(w->childHolderWidget, GTK_ALIGN_FILL);
+	gtk_box_set_homogeneous(GTK_BOX(w->childHolderWidget), TRUE);
```
cody271, line 386: What does this line do?

windows/draw.cpp
```diff
@@ -62,8 +62,16 @@ ID2D1HwndRenderTarget *makeHWNDRenderTarget(HWND hwnd)
 		&props,
 		&hprops,
 		&rt);
-	if (hr != S_OK)
-		logHRESULT(L"error creating HWND render target", hr);
+	if (hr != S_OK) {
```
cody271, line 65: Is the purpose of this addition to always try hardware acceleration *first*, then try software rendering fallback?
Needs a comment explaining this, if so.

windows/draw.cpp
```diff
@@ -509,3 +517,55 @@ void uiDrawRestore(uiDrawContext *c)
...
+
+	buf = uiprivNew(uiImageBuffer);
+
+	D2D1_BITMAP_PROPERTIES bp2 = D2D1::BitmapProperties();
+	bp2.dpiX = 0;
+	bp2.dpiY = 0;
+	bp2.pixelFormat = D2D1::PixelFormat(DXGI_FORMAT_R8G8B8A8_UNORM,
+				alpha ? D2D1_ALPHA_MODE_PREMULTIPLIED : D2D1_ALPHA_MODE_IGNORE);
+
+	//c->rt->BeginDraw();
```
cody271, line 537: Why are these `Begin()`/`End()` lines commented out?

matyalatte:
Removed unnecessary lines, fixed typos, and added a new enum `uiInterpMode`

```c
/**
 * Interpolation modes for image scaling.
 *
 * @enum uiInterpMode
 */
_UI_ENUM(uiInterpMode) {
	uiInterpModeSpeed,  //!< A high-performance interpolation. (Nearest neighbor on Windows and Unix.)
	uiInterpModeBalanced,  //!< A reasonable-performance interpolation. (Bilinear on Windows and Unix.)
	/**
	 * The highest-quality interpolation.
	 *
	 * @note This mode uses the bilinear interpolation on Windows.
	 */
	uiInterpModeQuality,
};
```

`uiInterpModeQuality` uses the bilinear interpolation on Windows because the cubic filter requires Windows 8 or later.
https://learn.microsoft.com/en-us/windows/win32/api/d2d1_1/ne-d2d1_1-d2d1_interpolation_mode

cody271:
@AndyObtiva I think you've been waiting on this PR for awhile now .. Needs some more testers. 😉

cody271:
> uiInterpModeQuality uses the bilinear interpolation on Windows because the cubic filter requires Windows 8 or later.
>
This needs to be a runtime check for Windows 8 or later, only *not* using high quality mode on legacy Windows as a fallback.

cody271:
A few more thoughts on `uiInterpMode` addition to the API:

> added a new enum `uiInterpMode`
>
> ```c
> _UI_ENUM(uiInterpMode) {
> 	...
> 	/**
> 	 * The highest-quality interpolation.
> 	 *
> 	 * @note This mode uses the bilinear interpolation on Windows.
> 	 */
> 	uiInterpModeQuality,
> };
> ```
>

Also I am not actually convinced that exposing different interpolation modes is relevant anymore. These sorts of image APIs are very old.. from a time when hardware was very slow, and exposing these kinds of low level details was necessary.

My first *guess* is that high quality mode by default will work fine! (On all platforms that support it)
Even pure software rendering 2D with Cairo/Direct2D/CoreGraphics, on older hardware, can probably render fast enough.

Have you measured the `uiImageBufferDraw()` performance difference of quality modes with something like your [libui-sprite-demo](https://github.com/matyalatte/libui-sprite-demo)?

Maybe always high quality really is too slow?
Then we could just add `uiImageBufferDrawFast()` with identical signature instead. No enum necessary.

matyalatte:
>> uiInterpModeQuality uses the bilinear interpolation on Windows because the cubic filter requires Windows 8 or later.
>
> This needs to be a runtime check for Windows 8 or later, only not using high quality mode on legacy Windows as a fallback.

It requires newer APIs, not only the enum for the cubic filter.
So, I'll leave it as it is with a TODO comment.

Old API (d2d1)
[ID2D1RenderTarget::DrawBitmap](https://learn.microsoft.com/en-us/windows/win32/direct2d/id2d1rendertarget-drawbitmap)
[D2D1_BITMAP_INTERPOLATION_MODE](https://learn.microsoft.com/en-us/windows/win32/api/d2d1/ne-d2d1-d2d1_bitmap_interpolation_mode)

Newer API (d2d1_1)
[ID2D1DeviceContext::DrawBitmap](https://learn.microsoft.com/en-us/windows/win32/api/d2d1_1/nf-d2d1_1-id2d1devicecontext-drawbitmap(id2d1bitmap_constd2d1_rect_f_float_d2d1_interpolation_mode_constd2d1_rect_f_constd2d1_matrix_4x4_f))
[D2D1_INTERPOLATION_MODE](https://learn.microsoft.com/en-us/windows/win32/api/d2d1_1/ne-d2d1_1-d2d1_interpolation_mode)

matyalatte:
> My first guess is that high quality mode by default will work fine!

Fair point.
I'll add `uiImageBufferDrawFast()` that uses the nearest neighbor, and remove the new enum.

> Have you measured the uiImageBufferDraw() performance difference of quality modes with something like your [libui-sprite-demo](https://github.com/matyalatte/libui-sprite-demo)?

No, I'll test it later.

darwin/draw.m
```diff
@@ -452,3 +452,97 @@ void uiDrawRestore(uiDrawContext *c)
...
+	return buf;
+}
+
+void uiImageBufferUpdate(uiImageBuffer *buf, const void *data)
+{
+	uint8_t *src = (uint8_t *) data;
+	uint8_t *dst = (uint8_t *) CGBitmapContextGetData(buf->buf);
+	int x, y;
+
+	// convert RGBA to BGRA and flip vertically
```
cody271, line 487: These hand written pixel format conversions are concerning..

Maybe a much larger performance problem than interpolation mode!

matyalatte: You mean, we don't need to consider the native-endian?
I just copied it from `uiImageAppend`, so Idk if it's required or not.
Should we fix `uiImageAppend` as well?

matyalatte:
Sorry, I noticed I didn't submit my comments. All of them were pending.

darwin/draw.m
matyalatte: Or you mean, I should use pointers like this?

```c
dst += buf->Stride * (buf->Height - 1);
for (y = 0; y < buf->Height; y++) {
    uint8_t *tmp_src = src;
    uint8_t *tmp_dst = dst;
    for (; tmp_dst < dst + buf->Stride; tmp_src += 4, tmp_dst += 4) {
        *tmp_dst = *(tmp_src + 2);
        *(tmp_dst + 1) = *(tmp_src + 1);
        *(tmp_dst + 2) = *tmp_src;
        *(tmp_dst + 3) = *(tmp_src + 3);
    }
    src += buf->Stride;
    dst -= buf->Stride;
}
```

cody271: > Or you mean, I should use pointers like this?
> 

No its a lot more complicated problem than that: image handling routines builtin to the underlying platforms are *heavily* optimized!

Native platforms are using GPU hardware acceleration, or at least SIMD assembly instructions for CPU fallback. This code here is just a very naive plain C loop. The best solution would be a pixel format all platforms can render natively. Then something simple like the Windows code might work on all platforms:

```cpp
void uiImageBufferUpdate(uiImageBuffer *buf, const void *data)
{
	D2D1_RECT_U rekt = D2D1::RectU(0, 0, buf->Width, buf->Height);
	buf->buf->CopyFromMemory(&rekt, data, buf->Stride);
}
```

cody271: > Darwin and unix require RGBA->BGRA conversion because they don't support RGBA.
> It will be a bottleneck when using uiImageBufferUpdate many times.
>
Okay I think we're saying the same thing here. 👍 

I'll have to do some more research on what's the best solution for all platforms.

matyalatte:
Removed `uiInterpMode` and added `uiImageBufferDrawFast()`.

And I made a tester app.
You can see a new binary `sprites_bench` when you build my demo.
https://github.com/matyalatte/libui-sprite-demo

It rotates 133x208 images with 2x upscaling.
You can change how many times it calls `uiImageBufferUpdate` per frame with the first spinbox.
And you can change how many times it calls `uiImageBufferDraw` per frame with the second spinbox.
You can also use `uiImageBufferDrawFast` if you check the checkbox.

https://github.com/libui-ng/libui-ng/assets/69258547/48707a6f-34f5-4296-9ff3-97b42b35c827

On my mid-end Windows PC, it calls `uiImageBufferDraw()` 10000 times per frame with 60fps.
And it calls `uiImageBufferUpdate()` 1000 times per frame with 60fps.
I also tried `uiImageBufferDrawFast` but I was not able to see any differences.

On my linux VM, it drops the frame rate just by calling `uiImageBufferDraw()` 10 times per frame.
And it's even heavier than `uiImageBufferUpdate`. (The cause might be my VM tho.)
But if you use `uiImageBufferDrawFast`, it can draw more than 100 images per frame with high frame rate.

Here are the release builds of the tester app.
You can try them for testing. (or you can build them with my demo by yourself)
- [sprites_bench-windows-release.zip](https://github.com/libui-ng/libui-ng/files/14297200/sprites_bench-windows-release.zip)
- [sprites_bench-ubuntu-release.zip](https://github.com/libui-ng/libui-ng/files/14297272/sprites_bench-ubuntu-release.zip)
- [sprites_bench-mac-x64-release.zip](https://github.com/libui-ng/libui-ng/files/14297321/sprites_bench-mac-x64-release.zip) (Note that this build doesn't support arm chips)

darwin/draw.m
matyalatte: I think it's ok to use BGRA in `uiImageBuffer`. (Someone might hate the inconsistency tho.)
We can add functions like `uiNewImageBufferFromFile` in the future.
Then, we don't need to care about the pixel format in most cases.

I also think it's ok to use the current implementation.
In most cases, we don't need to update buffers every frame.
My demo app calls `uiImageBufferUpdate` only in the first frame, and reuses the buffers in later frames.
And we can add faster functions for bit blitting in the future.

matyalatte: I noticed `CIImage` supports RGBA.
https://developer.apple.com/documentation/coreimage/ciformat?language=objc

And `vImage` supports RGBA<->BGRA conversion with `vImagePermuteChannels_ARGB8888()`.
https://developer.apple.com/documentation/accelerate/1533241-vimagepermutechannels_argb8888?language=objc

We can use either one and convert it to `CGImage`.

darwin/draw.m
```diff
@@ -452,3 +452,98 @@ void uiDrawRestore(uiDrawContext *c)
...
+}
+
+void uiImageBufferDraw(uiDrawContext *c, uiImageBuffer *buf, uiRect *srcrect, uiRect *dstrect)
+{
+	drawImageBuffer(c, buf, srcrect, dstrect, kCGInterpolationHigh);
+}
+
+void uiImageBufferDrawFast(uiDrawContext *c, uiImageBuffer *buf, uiRect *srcrect, uiRect *dstrect)
+{
+	drawImageBuffer(c, buf, srcrect, dstrect, kCGInterpolationLow);
```
matyalatte, line 541: I think I should use `kCGInterpolationNone` instead of `kCGInterpolationLow`.
I don't see any difference when I use `uiImageBufferDrawFast()` on my mac.
But it gains more fps when I replace `kCGInterpolationLow` with `kCGInterpolationNone`.

matyalatte:
I made a branch that uses [`vImage`](https://developer.apple.com/documentation/accelerate/vimage/) for RGBA <-> BGRA conversion.
https://github.com/matyalatte/libui-ng/commit/55746d9bb263f57b27ae5c1a5a40033414458c69

You can also use a branch of my demo for it.
https://github.com/matyalatte/libui-sprite-demo/tree/vimage

And here is the built binary from the branch.
[sprites_bench-vimage-mac-x64-release.zip](https://github.com/libui-ng/libui-ng/files/14313402/sprites_bench-vimage-mac-x64-release.zip)
(Note that this build doesn't support arm chips)

I thought it's faster than the native C loop but it's slower somehow.
I need someone can test it on the actual mac. I can't tell if the cause is the code or my VM.

cody271:
> I thought it's faster than the native C loop but it's slower somehow.
>
Your `Accelerate.framework` code looks correct, on first glance.

> I need someone can test it on the actual mac. I can't tell if the cause is the code or my VM.
>

Yes accurate graphics benchmarking and testing is not trivial!

Takes me a lot of time, as it requires many different machines both physical and virtual.

darwin/draw.m
cody271: > I think I should use `kCGInterpolationNone` instead of `kCGInterpolationLow`.
>
Yes should fully disable interpolation in "fast mode" where possible.

compile_log.txt
cody271, line 1: `compile_log.txt` ?

chnoblouch:
Is there anything specific left to be done here? This is the only missing feature I need to migrate my Qt app to libui-ng and I could help out by testing it on Windows 10/11 and arm64 macOS.

cody271:
> Is there anything specific left to be done here? I could help out by testing it
>
@chnoblouch Please do test this PR wherever you can! Thanks

chnoblouch:
Alright, I've run the QA executable on Windows, macOS, and Linux. Here are the screenshots:

**Windows 11 (x86-64):**


**macOS 14.3 (Apple M3):**


**Ubuntu 24.04 (x86-64, WSL):**


I've also run the sprites benchmark with and without `Accelerate` on macOS (Apple M3, compiled with Homebrew Clang 19). When calling `uiImageBufferUpdate` 1000 times per frame, I get ~44 FPS with the C loop version and ~48 FPS with the `Accelerate` version.

I was curious to find out why the difference wasn't bigger, so I took a look at the [assembly generated for the C loop version](https://godbolt.org/z/54jjT5Gxo). It seems like the compiler actually only slightly vectorizes the code. Maybe the overhead of calling `Accelerate` is just so high that you don't really notice the difference with small images?

cody271:
> I was curious to find out why the difference wasn't bigger, so I took a look at the [assembly generated for the C loop version](https://godbolt.org/z/54jjT5Gxo). It seems like the compiler actually only slightly vectorizes the code. Maybe the overhead of calling `Accelerate` is just so high that you don't really notice the difference with small images?

I have no idea, not my area really.. @trufae Any thoughts on this one?

cody271:
> Alright, I've run the QA executable on Windows, macOS, and Linux. Here are the screenshots:

Besides lack of DPI awareness in the Windows screenshot, looks good. See #294 

Also the QA page only has a single item. Someone will have to fill in the rest here, but it can be in a follow up PR.

trufae:
this is probably due a compiler optimization using vector instructions to access the buffer. The fix should be specifying to the function signature that those arguments must be aligned. In r2 i have this define to do that in a portable way

 ```
#if defined(_MSC_VER)
# define R_ALIGNED(x) __declspec(align(x))
#else
# define R_ALIGNED(x) __attribute__((aligned(x)))
#endif
```

I assume the alignment should be 16 for that case

chnoblouch:
I think I figured out why the compiler didn't optimize more aggressively. Due to C's aliasing rules, it probably had to assume that the pointer arguments could point to overlapping memory regions, so it wasn't allowed to vectorize the loop. I put a `restrict` qualifier on the pointers to mark them as non-aliased and now the C loop version runs at 55-60 FPS, beating even the `Accelerate` version! [Here's the resulting assembly](https://godbolt.org/z/dh4xz6q1e).

I also tried specifying the alignments as suggested by @trufae, but couldn't see any difference in the assembly. I might have just put the attributes in the wrong place though.

chnoblouch:
I'm also not even sure if `uiImageBufferUpdate` needs to be _that_ fast. Won't most applications load their images at startup, where performance is less important than in the main loop? Perhaps we could optimize this function later in a future PR?

chnoblouch:
@cody271 What is your opinion on this? Do you need benchmarks on other platforms before this PR can be merged?

rafael2k:
This is a really nice feature.

bcampbell:
For what it's worth, my use case is just to be able to directly edit pixels in raw memory and have a way to blit chunks of that to the screen as fast as possible.
I'd _love_ to ditch Qt on my pixel-painting app (http://evilpixie.scumways.com/) and use libui-ng instead.

petabyt:
You'd be looking for something like https://github.com/andlabs/libui/pull/386. Only support Windows/GTK though.

Issue #252, opened 2023-11-23
`uiWindowSetContentSize` may cause segmentation fault on Linux


kojix2:
Hi

I would like to thank the libui-ng team volunteers for continuing to create, review, and accept new pull requests this year.

We recently found that `uiWindowSetContentSize` may cause segmentation fault on Linux.

```c
#include <stdio.h>
#include <ui.h>

int onClosing(uiWindow *w, void *data) {
  uiQuit();
  return 1;
}

int main(void) {
  uiInitOptions o = {0};
  const char *err = uiInit(&o);
  
  if (err != NULL) {
    fprintf(stderr, "Error initializing libui-ng: %s\n", err);
    uiFreeInitError(err);
    return 1;
  }

  uiWindow *w = uiNewWindow("Hello World!", 300, 30, 0);
  uiWindowOnClosing(w, onClosing, NULL);
  uiLabel *l = uiNewLabel("Hello, World!");
  uiWindowSetChild(w, uiControl(l));
  uiControlShow(uiControl(w));

  int width, height;
  uiWindowContentSize(w, &width, &height);
  printf("Current window size: %d x %d\n", width, height);

  uiWindowSetContentSize(w, 500, 500);
  printf("Window size set to: 500 x 500\n");

  uiMain();
  uiUninit();
  return 0;
}
```

Could you please fix it so that uiWindowSetContentSize can be called even before uiMain and uiMainSteps are called on Linux?

---
This problem was first reported to [Glimmer DSL LibUI](https://github.com/AndyObtiva/glimmer-dsl-libui). 
See https://github.com/AndyObtiva/glimmer-dsl-libui/issues/61

Glimmer DSL for LibUI [workshop](https://rubyconf-2023.sessionize.com/session/531448) by @AndyObtiva was recently held at RubyConf2023 in San Diego. And it was reported that some of the [code](https://github.com/AndyObtiva/how-to-build-desktop-applications-in-ruby) used in the workshop caused segmentation violations in the case of Linux.

> In April, in libui-ng, pull request [#189 unix: Fix uiWindowContentSizeChanged()](https://github.com/libui-ng/libui-ng/pull/189) was merged. `uiWindowSetContentSize()` has been changed to call `uiMainStep` [here](https://github.com/libui-ng/libui-ng/blame/2746470ed8bae73588013f176b37baf32af42618/unix/window.c#L254).
> 
[...]

PR #261, opened 2023-12-10
uiWindow unit tests and fixes for unix

szanni:
This sadly got a lot bigger than initially anticipated but much of this code is interdependent, hence all of this at once: 

- Add unit tests for uiWindow
- Document uiWindowMargined default to be `FALSE`

Fixes for unix:
- Fix uiWindowContentSize upon window creation, see #252
- Make uiWindowSetPosition non blocking when hidden
- Make uiWindowSetContentSize non blocking when hidden
- Fix uiWindowSetContentSize for margined windows

`uiWindowContentSize` now uses  `gtk_window_get_size` and `gtk_window_resize` instead of trying to do some magic calculations on the inner container. According to the docs gtk already does a best effort calculation to report the size sans window decorations. The prior, inner container size code does not work on hidden windows and margined windows anyways and does not add anything on all systems I tested (i3wm and weston) apart from code, confusion, bugs and brittleness, hence the removal.

It is unclear to me if there was ever a good reason for doing these calculations. I could not find any documentation in the code or git log. All code related to window sizing is a best effort on all Unix systems anyways as the [gtk3 docs](https://docs.gtk.org/gtk3/method.Window.get_size.html) state everywhere. I did not notice any regressions by removing all the window size calculations.

### DIFF
--- a/unix/main.c
+++ b/unix/main.c
@@ -19,6 +19,10 @@ const char *uiInit(uiInitOptions *o)
 	uiprivInitAlloc();
 	uiprivLoadFutures();
 	timers = g_hash_table_new(g_direct_hash, g_direct_equal);
+
+	// Run the event loop manually by default to ensure we can run uiMainStep()
+	// internally to make asynchronous GTK calls appear synchronous.
+	uiMainSteps();
 	return NULL;
 }
 
--- a/test/unit/main.c
+++ b/test/unit/main.c
@@ -27,6 +27,7 @@ int unitTestSetup(void **_state)
 	uiInitOptions o = {0};
 
 	assert_null(uiInit(&o));
+	state->c = NULL;
 	state->w = uiNewWindow(UNIT_TEST_WINDOW_TITLE, UNIT_TEST_WINDOW_WIDTH, UNIT_TEST_WINDOW_HEIGHT, 0);
 	uiWindowOnClosing(state->w, unitWindowOnClosingQuit, NULL);
 	return 0;
@@ -36,7 +37,8 @@ int unitTestTeardown(void **_state)
 {
 	struct state *state = *_state;
 
-	uiWindowSetChild(state->w, uiControl(state->c));
+	if (state->c != NULL)
+		uiWindowSetChild(state->w, uiControl(state->c));
 	uiControlShow(uiControl(state->w));
 	//uiMain();
 	uiMainSteps();
--- a/test/unit/main.c
+++ b/test/unit/main.c
@@ -59,6 +59,7 @@ int main(void)
 	int failedComponents = 0;
 	struct unitTest unitTests[] = {
 		{ initRunUnitTests },
+		{ windowRunUnitTests },
 		{ menuRunUnitTests },
 		{ sliderRunUnitTests },
 		{ spinboxRunUnitTests },
--- a/test/unit/main.c
+++ b/test/unit/main.c
@@ -27,7 +27,7 @@ int unitTestSetup(void **_state)
 	uiInitOptions o = {0};
 
 	assert_null(uiInit(&o));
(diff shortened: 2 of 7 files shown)


unix/main.c
```diff
@@ -19,6 +19,10 @@ const char *uiInit(uiInitOptions *o)
 	uiprivInitAlloc();
 	uiprivLoadFutures();
 	timers = g_hash_table_new(g_direct_hash, g_direct_equal);
+
+	// Run the event loop manually by default to ensure we can run uiMainStep()
+	// internally to make asynchronous GTK calls appear synchronous.
+	uiMainSteps();
```
cody271, line 25: Shouldn't the event loop default to "steps mode" on Darwin now as well?

cody271 (commented):
Has this been tested on Wayland yet?

cody271:
> It is unclear to me if there was ever a good reason for doing these calculations. I could not find any documentation in the code or git log. All code related to window sizing is a best effort on all Unix systems anyways as the [gtk3 docs](https://docs.gtk.org/gtk3/method.Window.get_size.html) state everywhere. I did not notice any regressions by removing all the window size calculations.
>
👍 

>> on all systems I tested (i3wm and weston)
>>
> Has this been tested on Wayland yet?

I will try some other DEs besides Weston.

unix/main.c
szanni: > Shouldn't the event loop default to "steps mode" on Darwin now as well?

I guess that would make sense for consistencies sake? On the other hand Windows has been doing it's own thing all along, which has been calling `uiMainStep()` in `uiMain()`.

Happy to change that, just needs to be tested very thoroughly to make sure we don't break anything. There seems to be some interaction around that and uiMenu.

szanni:
> > > on all systems I tested (i3wm and weston)
> > 
> > Has this been tested on Wayland yet?
> 
> I will try some other DEs besides Weston.

To add to that, this means the unit test work. By no means do Wayland systems honor `uiWindowSetPosition` in any shape or form as Wayland [has no protocol for doing so](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/247). And SetContentSize only really works when increasing window size form the original on Wayland systems consistently.
Tested with weston and sway+floating windows.

kojix2:
Hi szanni.
Thank you for creating this pull request. Even before merging this pull request into libui-ng/master, our Ruby team can distribute the pre-package that includes it.

In this way, if any new problems occur, there is a good chance that someone will find and report them.
If you don't mind, we would be grateful if you could resolve the conflicts. Thank you.

cody271:
> Tested with weston and sway+floating windows.
>

Confirmed tests passing on both GNOME/Wayland and Weston:

```
[==========] Running 11 test(s).
[ RUN      ] windowNew
[       OK ] windowNew
[ RUN      ] (windowDefaultMargined)
[       OK ] (windowDefaultMargined)
[ RUN      ] (windowMargined)
[       OK ] (windowMargined)
[ RUN      ] (windowDefaultTitle)
[       OK ] (windowDefaultTitle)
[ RUN      ] (windowDefaultContentSize)
[       OK ] (windowDefaultContentSize)
[ RUN      ] (windowSetTitle)
[       OK ] (windowSetTitle)
[ RUN      ] (windowSetPosition)
[       OK ] (windowSetPosition)
[ RUN      ] (windowSetContentSize)
[       OK ] (windowSetContentSize)
[ RUN      ] (windowMarginedSetContentSize)
[       OK ] (windowMarginedSetContentSize)
[ RUN      ] (windowSetContentSizeNoCallback)
[       OK ] (windowSetContentSizeNoCallback)
[ RUN      ] (windowSetPositionNoCallback)
[       OK ] (windowSetPositionNoCallback)
[==========] 11 test(s) run.
[  PASSED  ] 11 test(s).
```

cody271 (approved):
LGTM

PR #262, opened 2023-12-13
darwin: Fix Darwin builds by removing `soversion`.

### LINKED ISSUE #258: Meson 1.3.0 only accepts numeric versions as `soversion` on macOS
You can't build libui-ng with meson 1.3.0 on macOS, becuase it uses `soversion` for `-compatibility_version` and `-current_version`. (Idk if it's a bug or not tho.)

```
cc  -o ***.A.dylib ***.o ... -compatibility_version A -current_version A ...

ld: malformed 32-bit x.y.z version number: A
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

To fix the error, we should use a numeric verison like `0`, or remove `soversion` and `darwin_versions` from library() options.

szanni:
Meson 1.3.0 does not accept 'A' as the soversion anymore which breaks Darwin shared library builds.
Remove soversion versioning from all platforms as tagged API versions are lacking from the library at the moment anyways.

Fixes #258 (and our CI runners for macOS)

### DIFF
--- a/darwin/meson.build
+++ b/darwin/meson.build
@@ -60,6 +60,5 @@ libui_deps += [
 		modules: ['Foundation', 'AppKit'],
 		required: true),
 ]
-libui_soversion = 'A'
 # the / is required by some older versions of OS X
 libui_rpath = '@executable_path/'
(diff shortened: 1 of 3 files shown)


cody271 (approved):
LGTM

PR #50, opened 2022-01-25
[Linux+Windows] allow dialogs without a parent window.

msink:
Part of andlabs/libui#375
- untested
- macOS missed

PR #267, opened 2024-01-27
Allow dialogs without a parent window

matyalatte:
This is revival of https://github.com/libui-ng/libui-ng/pull/50.

I added support for macOS.

And tested the following code on Windows10, Ubuntu20.04, and macOS 10.15.
```c
int main(void)
{
    uiInitOptions options;
    const char *err;

    memset(&options, 0, sizeof (uiInitOptions));
    err = uiInit(&options);
    if (err != NULL) {
        fprintf(stderr, "error initializing libui: %s", err);
        uiFreeInitError(err);
        return 1;
    }

    uiMsgBox(NULL, "NULL Parent Test", "Dialogs wihtout a parent window.");
}
```

This feature is useful because you can use dialogs before creating the main window.
For example, you can close your app with an error dialog before the main loop.
I hope it will be merged soon.

### DIFF
--- a/darwin/stddialogs.m
+++ b/darwin/stddialogs.m
@@ -5,7 +5,7 @@
 // LONGTERM explicitly document this works as we want
 // LONGTERM note that font and color buttons also do this
 
-#define windowWindow(w) ((NSWindow *) uiControlHandle(uiControl(w)))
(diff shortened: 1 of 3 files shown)


cody271 (approved):
LGTM

PR #271, opened 2024-03-24
windows: Fix time string length retrieval

kojix2:
Hello.

I was thinking about how I can contribute to the project easily and without any difficulty, and I came up with the idea of having ChatGPT review libui-ng.

According to ChatGPT, there is a mistake here.
It may indeed be so.

### DIFF
--- a/windows/datetimepicker.cpp
+++ b/windows/datetimepicker.cpp
@@ -83,7 +83,7 @@ static void setDateTimeFormat(HWND hwnd)
 	uiprivFree(unexpandedDate);
 
 	ntime = GLI(LOCALE_STIMEFORMAT, NULL, 0);
-	if (ndate == 0)
+	if (ntime == 0)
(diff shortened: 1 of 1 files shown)


szanni:
Looks like a legitimate bug to me.

The `GetLocaleInfoEx` docs sadly do not state returned string on error is actually `\0` terminated, so unsure if this would crash nonetheless on `datetime = strf(L"%s %s", date, time);`. But this patch definitely is an improvement and possibly all that is needed.

cody271:
> Looks like a legitimate bug to me.

Confirmed bug, checks `ndate` for error result twice instead of `ntime`.

cody271 (approved):
LGTM

rubyFeedback:
Interesting. I did not know ChatGPT can be used for finding bugs. Can this also be applied to e. g. a codebase in (pure) ruby?

PR #272, opened 2024-03-24
windows: Fix separator orientation

kojix2:
Hello.

I was thinking about how I can contribute to the project easily and without any difficulty, and I came up with the idea of having ChatGPT review libui-ng.

According to ChatGPT, it could be wrong here.

> In the `uiSeparator *uiNewVerticalSeparator(void)` function, despite creating a new separator that is vertical, `SS_ETCHEDHORZ` is being passed as an argument. This constant represents a horizontal line style. Therefore, `SS_ETCHEDVERT` should be used here instead.

- https://learn.microsoft.com/en-us/windows/win32/controls/static-control-styles

However, I am not sure. It is possible that the code is intentional. If there is no problem, please close it.

### DIFF
--- a/windows/separator.cpp
+++ b/windows/separator.cpp
@@ -64,7 +64,7 @@ uiSeparator *uiNewVerticalSeparator(void)
 
 	s->hwnd = uiWindowsEnsureCreateControlHWND(0,
 		L"static", L"",
-		SS_ETCHEDHORZ,
+		SS_ETCHEDVERT,
 		hInstance, NULL,
 		TRUE);
 	s->vertical = TRUE;


cody271:
Duplicate of #234?

szanni:
> https://learn.microsoft.com/en-us/windows/win32/controls/static-control-styles

`SS_ETCHEDHORZ: Draws the top and bottom edges of the static control using the EDGE_ETCHED edge style.`
`SS_ETCHEDVERT: Draws the left and right edges of the static control using the EDGE_ETCHED edge style.`

From `ui.h`:
`uiNewHorizontalSeparator: Creates a new horizontal separator to separate controls being stacked vertically.`
`uiNewVerticalSeparator: Creates a new vertical separator to separate controls being stacked horizontally.`

You can test it with an old test case I seemingly never submitted #273.

Technically this patch is correct. It seemingly makes no difference on windows though from what I can tell.

Without and without patch:
image

szanni (approved):
Seemingly technically correct with no observable effect on the systems I've tested though.

cody271:
> Seemingly technically correct with no observable effect on the systems I've tested though.

Patch is correct but confirmed that doesn't seem to matter here on Windows 11.

cody271 (approved):
LGTM

PR #279, opened 2024-05-18
Update macOS runner

matyalatte:
Github Actions will drop `macos-11` support on 6/28/2024.

- [About GitHub-hosted runners - GitHub Docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)
- [GitHub Actions: macOS 14 (Sonoma) is now available - The GitHub Blog](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/)

So, we have to use a newer version of macOS runners.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -146,8 +146,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # macos-latest (macos-10.15), macos-11
-        os: [macos-11]
+        os: [macos-latest]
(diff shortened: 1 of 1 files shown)


cody271:
> So, we have to use a newer version of macOS runners.
>
Also this way the CI jobs all use `-latest` OS images, including macOS.

cody271 (approved):
LGTM

PR #280, opened 2024-06-01
Initial uiScroll for windows

petabyt:
- Do we all agree on `uiNewScroll` and `uiScrollSetChild`?
- Do we handle horizontal scrolling?
- Does this implementation suck?
- (Most of this implementation is copied from other libui code)
- We should probably get GTK and Cocoa uiScroll working first before this is merged

I haven't added any tests or examples yet but it works as you'd expect:
image

PR #281, opened 2024-06-04
uiScroll for unix and darwin

swissChili:
Adding to #280, this PR implements uiScroll for unix and darwin with the same API.

darwin

unix


petabyt:
Nice!
image

swissChili:
Something should probably be done about the horizontal scrolling. For each axis, we want either: a scroll bar, an overflow, or the child to be constrained to the parent size.

Maybe something like this?

```c
enum {
    UI_SCROLL_BAR, UI_SCROLL_CLIP, UI_SCROLL_SQUEEZE
}

void uiScrollSetPolicy(uiScroll *scroll, int horizontal, int vertical);
```

petabyt:
> Something should probably be done about the horizontal scrolling. For each axis, we want either: a scroll bar, an overflow, or the child to be constrained to the parent size.
> 
> Maybe something like this?
> 
> ```c
> enum {
>     UI_SCROLL_BAR, UI_SCROLL_CLIP, UI_SCROLL_SQUEEZE
> };
> 
> void uiScrollSetPolicy(uiScroll *scroll, int horizontal, int vertical);
> ```
> 
> So for your example you would probably want `UI_SCROLL_SQUEEZE, UI_SCROLL_BAR` so the text would flow as expected.

Those enums sound familiar.. CSS? :)
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x

So for `UI_SCROLL_SQUEEZE`, does that mean the window size of my constitution app would stretch to the size of the longest label? And of course for buttons/boxes it would stretch normally.
On that note, we probably should implement word wrap, but I guess that's another PR.

swissChili:
I more meant that SQUEEZE would force the child to be the width of the parent, which can be anything. I guess we actually need four settings:
- parent and child can be any size, child is clipped
- parent and child can be any size, child scrolls
- parent can be any size, child must be size of parent
- parent must be size of child, child can be any size

For a long document, you would want the child width to be fixed to the parent width (so that lines must break), but have potentially larger height and be able to scroll vertically.

swissChili:
I'm also not sure how to implement this in the various toolkits. On Mac it's easy: you can specify all this with constraints. With GTK you can set the max/min content width/height on a scroll view, and toggle the scrollbar visibility accordingly—I think that would work. No idea about windows.

petabyt:
> For a long document, you would want the child width to be fixed to the parent width (so that lines must break), but have potentially larger height and be able to scroll vertically.

I don't think we currently have the ability to do line break. It might be an easy refactor on GTK or Cocoa but I think on Windows it would need a new widget (`uiWrappingLabel`) or just an option (`uiLabelSetWrapping`)uiNewMultilineEntry. It shouldn't be that hard though, at least I hope.

> parent can be any size, child must be size of parent

So without text wrap, a long label would just clip. With text wrapping we shouldn't need this option, the sizing of a wrapped label would just behave like a box.

swissChili:
> So without text wrap, a long label would just clip. With text wrapping we shouldn't need this option, the sizing of a wrapped label would just behave like a box.

You're right, and it's not very useful in this case. But if you had e.g. a scrolling list of images, you might want each image to be the width of the container. I'm not sure if this is possible on Windows with the way it's currently implemented, but this behavior is easy on mac and GTK.

petabyt:
> but if you wanted to force their width, you would need a different option

Hmm this would need image scaling so it would have to be reasonably fast. Scrolling is already a little glitchy on Windows. I think for an image you could just specify a minimum size in px or dp and the sizing of the image frame (or set it to full res) would behave similarly to a button.
I played with GTK images in libui and this was basically what the behavior was like.

petabyt:
Here's your scrolling example on Windows:
image

If it's easy to get clip/squeeze working on GTK/Cocoa, then you can go ahead and implement that. It'd help me to have a working uiScroll as a reference when getting the win32 implementation working.

petabyt:
Ok here's a codepen, do you think this will cover most use cases?
https://codepen.io/petabyt/pen/OJYxNZO?editors=1100

petabyt:
@swissChili Any comments?

swissChili:
I haven't had time to work on this recently, but that looks good. It might even be overkill, for a first revision I'll probably just do horizontal/vertical scroll + fit in the other direction.

cblc:
How is this progressing? Is there hope to have it merged soon together with the Windows-version PR, or are there still decisions to be made?

petabyt:
The code is mostly finished but needs some more review/suggestions, so far nobody has offered any. So until then this won't be merged.

petabyt:
Any comment from @cody271 would be nice

cody271:
Hi @petabyt! So my main concern here is the horizontal scrolling.

> Maybe something like this?
> > ```c
> > enum {
> >     UI_SCROLL_BAR, UI_SCROLL_CLIP, UI_SCROLL_SQUEEZE
> > };
> > 
> > void uiScrollSetPolicy(uiScroll *scroll, int horizontal, int vertical);
> > ```
> > 
>

`uiScrollSetPolicy()` needs to match [`gtk_scrolled_window_set_policy()`](https://docs.gtk.org/gtk3/method.ScrolledWindow.set_policy.html) and [GtkPolicyType](https://docs.gtk.org/gtk3/enum.PolicyType.html) more closely.
That way scrolling policy behavior is consistent on all platforms. Also not sure how much this differs from the CSS spec?

common/controlsigs.h
```diff
@@ -19,6 +19,7 @@
 #define uiMultilineEntrySignature 0x4D6C6E45
 #define uiProgressBarSignature 0x50426172
 #define uiRadioButtonsSignature 0x5264696F
+#define uiScrollSignature 0x5dcd9b83
```
cody271, line 22:
```suggestion
#define uiScrollSignature 0x5363726c
```

petabyt:
> So my main concern here is the horizontal scrolling.

Is there any difference between horizontal and vertical scrolling? We want the same behavior for both directions right?

Did you see this? https://codepen.io/petabyt/pen/OJYxNZO?editors=1100  
I think it should cover every use case.

cody271:
> We want the same behavior for both directions right?
>
Same behavior I think, yes. Single function to set both policies.

> Did you see this? https://codepen.io/petabyt/pen/OJYxNZO?editors=1100
>
If I'm reading that correctly, would map to something like this for `GtkScrolledWindow`?
```c
Clip = GTK_POLICY_EXTERNAL
Fit = GTK_POLICY_NEVER
Scroll = GTK_POLICY_ALWAYS
```

petabyt:
I can try and get clip/fit/scroll working on Windows. For darwin I have no mac to test on, so I can't help there. I don't know if @swissChili can finish it or do you @cody271 want to work on it?

PR #289, opened 2024-08-06
Add uiTabOnSelected for all platforms

### LINKED ISSUE #253: Tab selection events
Hi,

In the current LibUI API, I did not find any way to listen to the event of when a user selected a certain tab to see its content. 

It would be great to have that functionality as it would enable things like:
- Tracking which tab the user has selected to perform certain actions when the tab content is displayed, like remembering tab browsing history or refreshing data.
- Delaying the population of tab content until a tab is selected, which can optimize the startup time of certain apps.

For an example of such an implementation, Eclipse SWT `TabFolder` supports tracking tab selection with the `addSelectionListener` method:
https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/swt/widgets/TabFolder.html

I use that functionality in [Gladiator (Glimmer Editor)](https://github.com/AndyObtiva/glimmer-cs-gladiator) to save a history of user tab selection actions to remember which files the user viewed last.

It would be great if I could have that same functionality in LibUI. Initially, the "tab selected event" could be sufficient. Eventually, the "tab unselected event" might be useful too (like for performing cleaning of resources).

mohad12211:
closes #253 

- [x] add documentation to ui.h. 
- [x] compare how `_UI_EXTERN void uiTabSetSelected(uiTab *t, int index);` behaves when `index` is out of bounds and what should be done.

### DIFF
--- a/test/page1.c
+++ b/test/page1.c
@@ -60,9 +60,14 @@ SHED(disable, Disable)
 
 uiBox *page1;
 
+void goToPage2(uiButton *b, void *data)
+{
+	uiTabSetSelected(mainTab, 1);
+}
+
 void makePage1(uiWindow *w)
 {
-	uiButton *getButton, *setButton;
+	uiButton *getButton, *setButton, *nextButton;
 	uiBox *hbox;
 	uiBox *testBox;
 	uiLabel *label;
@@ -167,5 +172,8 @@ void makePage1(uiWindow *w)
 	uiBoxAppend(hbox, uiControl(getButton), 1);
 	uiBoxAppend(page1, uiControl(hbox), 0);
 
+	nextButton = uiNewButton("Go to Page 2");
+	uiButtonOnClicked(nextButton, goToPage2, NULL);
+	uiBoxAppend(page1, uiControl(nextButton), 0);
 	uiBoxAppend(page1, uiControl(label), 0);
 }
--- a/ui.h
+++ b/ui.h
@@ -915,7 +915,12 @@ _UI_EXTERN uiLabel *uiNewLabel(const char *text);
 typedef struct uiTab uiTab;
 #define uiTab(this) ((uiTab *) (this))
 
-_UI_EXTERN void uiTabOnSelected(uiTab *t, void (*f)(uiTab *, int, void *), void *data);
+_UI_EXTERN int uiTabSelected(uiTab *t);
+
+_UI_EXTERN void uiTabSetSelected(uiTab *t, int index);
+
+_UI_EXTERN void uiTabOnSelected(uiTab *t,
+        void (*f)(uiTab *sender, void *senderData), void *data);
 
 /**
  * Appends a control in form of a page/tab with label.
--- a/ui.h
+++ b/ui.h
@@ -915,10 +915,37 @@ _UI_EXTERN uiLabel *uiNewLabel(const char *text);
 typedef struct uiTab uiTab;
 #define uiTab(this) ((uiTab *) (this))
 
+/**
+ * Returns the index of the tab selected.
+ *
+ * @param c uiTab instance.
+ * @returns Index of the tab selected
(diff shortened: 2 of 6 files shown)


mohad12211:
- `uiTabOnSelected(t, defaultOnSelected, NULL);` is called before destroying the `uiTab`, because if tab index 0 is currently selected, and it's destroyed, focus will shift into tab index 1, which will trigger a selected event.
- appending/inserting after `uiTabOnSelected` might trigger an event call if the selected page changes, even though the user hasn't actually changed the selection, I think this is the desired behaviour.

- [ ] not documentation added to ui.h yet.

ui.h
```diff
@@ -915,6 +915,8 @@ _UI_EXTERN uiLabel *uiNewLabel(const char *text);
 typedef struct uiTab uiTab;
 #define uiTab(this) ((uiTab *) (this))
 
+_UI_EXTERN void uiTabOnSelected(uiTab *t, void (*f)(uiTab *, int, void *), void *data);
```
cody271, line 918: This should match the existing selection APIs (combo box and radio button) more closely:
```c
_UI_EXTERN int uiTabSelected(uiTab *t);

_UI_EXTERN void uiTabSetSelected(uiTab *t, int index);

_UI_EXTERN void uiTabOnSelected(uiTab *t,
        void (*f)(uiTab *sender, void *senderData), void *data);
```

mohad12211:
made it similar to combo box and radio button selection API, I should test and research how different platforms handles negative and out of bounds `index` at `uiTabSetSelected` and what should we do. any suggestions?

cody271:
> made it similar to combo box and radio button selection API, I should test and research how different platforms handles negative and out of bounds `index` at `uiTabSetSelected` and what should we do. any suggestions?

Currently for negatives `-1` is a special value for "no selection", and out of bounds requests just fail silently.

test/page1.c
```diff
@@ -167,5 +172,8 @@ void makePage1(uiWindow *w)
 	uiBoxAppend(hbox, uiControl(getButton), 1);
 	uiBoxAppend(page1, uiControl(hbox), 0);
 
+	nextButton = uiNewButton("Go to Page 2");
```
cody271, line 175: Broken rendering of the tab on Windows when I click `Go to Page 2`:

tab_select

Renders correctly if I manually select the `Page 2` tab _first_.

mohad12211: thanks for testing, I thought it was my `wine` doing something weird.

mohad12211: should be fixed

mohad12211:
> Currently for negatives `-1` is a special value for "no selection"

not sure if "no selection" makes sense for `uiTab`, I would suggest making `-1` also a silent failure.

cody271 (commented):
> not sure if "no selection" makes sense for `uiTab`, I would suggest making `-1` also a silent failure.
>
Just tested this again,`uiTabSetSelected(mainTab, -1)` currently does something slightly different on each platform:
On Unix, GTK interprets this `-1` index as the "last page" and shows `Page 5`. Windows "deselects" the tab page entirely. Only Darwin implementation does the right thing: should be silent failure for _any_ invalid index. Otherwise looks good.

mohad12211:
Done. Ready to be merged.

cody271:
> Done. Ready to be merged.
>
Yeah just copy pasting the same bounds check is fine here.  👍 

> > Currently for negatives `-1` is a special value for "no selection"
> 
> not sure if "no selection" makes sense for `uiTab`, I would suggest making `-1` also a silent failure.
>
Also confirmed that `uiTabSelected(uiNewTab()) == -1` always by default. So while this doesn't really make any _logical_ sense, you are correct.. At least now tab selection values are totally consistent on all platforms, both read/write.

cody271 (approved):
LGTM

Issue #290, opened 2024-08-07
Why is the Windows static library named libui.a?


kojix2:
Hi! I have a basic question today.

Recently, I've been working on creating a Crystal binding for libui-ng. The Ruby binding uses FFI to call libui-ng functions from a shared library, while Crystal links a static library to build an executable.

I don't know much about static linking with MSVC on Windows, so I'm working with ChatGPT to learn how. Then I noticed something strange. When I download the static library for Windows created by GitHub Actions, it is named "libui.a".

According to ChatGPT (Kojix2 doesn't have anyone nearby to consult about Windows)  static libraries for Windows are usually named "libui.lib", and the ".a" extension is unusual. In fact, if I change the file name to "libui.lib", it compiles successfully. Why is that?

Since libui-ng is a cross-platform library, does it follow the Linux or Mac naming convention?
I don't think that's the case, but if it is a typo, does it need to be fixed?

petabyt:
I avoid MSVC at all costs, but with mingw the .a links just fine, same as GCC

kojix2:
Thank you for your clarification.

I agree that using MinGW can be a good idea. However, the Windows version of Crystal only supports MSVC, and my project needs to use MSVC because Crystal on Windows links to MSVC libraries.

It seems like a good idea to keep the static library extension as `.a` for MinGW and change it to `.lib` for MSVC. 
This should help avoid confusion and ensure compatibility.

matyalatte:
I found a related issue in the meson project.
https://github.com/mesonbuild/meson/issues/8153

And [the document](https://mesonbuild.com/Reference-manual_functions.html#library_name_suffix) says this behavior is intentional.

> By convention MSVC static libraries use the lib suffix, but we use a to avoid a potential name clash with shared libraries which also generate import libraries with a lib suffix.

I'm not sure what they exactly mean tho. It means you can merge #291 if libui does not build import libraries, I guess?

kojix2:
@petabyt 
Finally the Crystal language has come to MinGW. 
Now there is no reason to use MSVC. From now on I will avoid MSVC at all costs!

PR #291, opened 2024-08-08
Rename static library suffix for Windows MSVC

kojix2:
Change the static library extension to .lib for MSVC.

See #290

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -125,19 +125,9 @@ add_project_arguments(libui_project_compile_args,
 add_project_link_arguments(libui_project_link_args,
 	language: ['c', 'cpp', 'objc'])
 
-if libui_OS == 'windows'
-	if libui_MSVC
-		static_lib_suffix = 'lib'
-	else
-		static_lib_suffix = 'a'
-	endif
-	shared_lib_suffix = 'dll'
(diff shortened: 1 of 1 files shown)


petabyt:
LGTM @cody271

meson.build
```diff
@@ -125,6 +125,21 @@ add_project_arguments(libui_project_compile_args,
...
+	endif
+	shared_lib_suffix = 'dll'
+elif libui_OS == 'darwin'
+	static_lib_suffix = 'a'
+	shared_lib_suffix = 'dylib'
+else
+	static_lib_suffix = 'a'
+	shared_lib_suffix = 'so'
+endif
+
```
matyalatte, line 142: You can use `[]` for the default behavior.
```suggestion
libui_suffix = []
if libui_OS == 'windows' and libui_MSVC
	libui_suffix = 'lib'
endif

```

cody271 (changes requested):
This doesn't work? Builds immediately fail with:
```
meson.build:170:41: ERROR: Unknown variable "static_lib_suffix".
```

kojix2:
There are many other issues that may be more important.

PR #295, opened 2024-09-12
Fix issues related to modal windows on macOS

### LINKED ISSUE #288: Mac OS menu bar items unclickable after modal dialog popup.
I am maintaining the Rust wrapper of libui-ng and [one of my users reported](https://github.com/libui-rs/libui/issues/8) that the menu bar items under MacOS become _unclickable_
after a _modal dialog popup closed_. This has been confirmed by another user of mine. Because I don't have a Mac, I can only do the project management side however, and report this here.

### Steps to reproduce:

1. Create an application that opens a `uiMsgBox` (or `uiOpenFolder`, `uiOpenFile`, ..) from a menu item. Preferrably a system menu item as created with `uiMenuAppendAboutItem` for example. Then run the app.
2. Click the about button. The window should loose focus and the message box should appear in the middle of the window.
3. Close the message box, but don't click anywhere else!
4. Click the about button again. The message box now appears in the middle of the screen (not window!).
5. Close the message box.
6. Try clicking the about button a third time. Observe that the menu items are now locked.

**Here is a video showing this procedure (altough with two different menu items, same effect):**

https://github.com/libui-rs/libui/assets/36557882/28bee401-0242-4870-a51a-69a52923d68d

### Possible solution or workaround:

Another user noted that:
> _"As long as I refocus the window in between menu items, I can repeat it as many times as I want (as far as I can tell), and the bug never occurs."_

Indeed, calling `uiControlShow()` on the window, right after the message box closes forces a focus of the window and the bug doesn't appear.

### Next steps?
There might be some issue with the window focus that requires investigation by someone with a Mac on knowledge of the Cocoa framework (@cody271, @matyalatte?). I can't imagine that manually retaking the focus is the correct way here.

I hope this can be fixed in the Mac portion of the code in libui-ng, rather than downstream in the rust wrapper :)

matyalatte:
- Fixed a bug where `keyWindow` could be nil after calling modal windows with [libui-rs](https://github.com/libui-rs/libui).
- Fixed a bug where menu items became unclickable when using a null pointer as the parent of a modal window.

fixes #288, and maybe you can close #164 as well.


darwin/stddialogs.m
```diff
@@ -124,6 +136,11 @@ static void msgbox(NSWindow *parent, const char *title, const char *description,
 	[cm run];
 	[cm release];
 	[a release];
+
+	// Explicitly set a window as the key window, as it can be nil under certain circumstances.
+	// (https://github.com/libui-ng/libui-ng/issues/288)
+	if (parent)
+		[parent makeKeyWindow];
```
matyalatte, line 143: I noticed that we should check if the parent is `keyWindow` or not. Or we can move this line to [uiprivMenuItem](https://github.com/libui-ng/libui-ng/blob/49b04c4cf8ae4d3e38e389f446ef75170eb62762/darwin/menu.m#L74) if `keyWindow` can not be nil without `uiMenuItemOnClicked`.

trufae:
any plans to merge this and any of the other 42 PRs?

cody271:
> any plans to merge this and any of the other 42 PRs?

Looking at this one now

cody271 (commented):
> * Fixed a bug where `keyWindow` could be nil after calling modal windows with [libui-rs](https://github.com/libui-rs/libui).
>
This commit should be moved to its own PR. I can't reproduce in C or Go.
Seems to be a Rust bindings specific bug, needs further investigation.

> * Fixed a bug where menu items became unclickable when using a null pointer as the parent of a modal window.
> 
Confirmed this fixes the menu with parentless modal windows, even without `keyWindow` commit.

PR #297, opened 2024-10-09
The first window displayed center, second and subsequent displayed ca…

junogura:
Change Initial position of uiWindow on macOS.

1. The First Window is displayed at center.
2. Second and subsequent are displayed cascaded.

Because (single)App and DocumentApp templates of SwiftUI do like this.

### DIFF
--- a/darwin/window.m
+++ b/darwin/window.m
@@ -170,7 +170,15 @@ static int uiWindowVisible(uiControl *c)
 
 static void uiWindowShow(uiControl *c)
 {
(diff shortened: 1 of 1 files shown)


cody271:
> 
> Because (single)App and DocumentApp templates of SwiftUI do like this.

Hi @junogura, yes that's a much better default. Looks good! Will need more testing with multiple window apps.

cody271:
Replaces #100 ?

PR #308, opened 2025-01-26
uiInitOptions NULL

pawplo:
Allow uiInitOptions to be set to NULL.

### DIFF
--- a/darwin/main.m
+++ b/darwin/main.m
@@ -113,7 +113,9 @@ uiInitOptions uiprivOptions;
 const char *uiInit(uiInitOptions *o)
 {
 	@autoreleasepool {
-		uiprivOptions = *o;
+		if (o) {
+			uiprivOptions = *o;
+		}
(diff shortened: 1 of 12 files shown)


cody271 (commented):
Overall I like this idea- allowing `NULL` to mean "all options set to their default values" isn't really a breaking change to the API and will make life much easier for language bindings that currently have to deal with empty `uiInitOptions` structs. Although let's keep this as small as possible and save the changes to the examples for a later PR.

trufae:
I think it will be good if public apis specify which parameters are nullable and which are not, at least gcc and clang have attributes for this, and this can help the code gen and checks to be made at compile time to spot bugs without the need to add assert guards and such. For example:

* https://github.com/google-deepmind/mujoco/issues/309
* https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html (^F nonnull)

PR #313, opened 2025-03-29
Add libui-vala to bindings in README.md

ibinshoid:


kojix2:
It's kind of interesting that vala doesn't use GTK and uses libui.

ibinshoid:
> It's kind of interesting that vala doesn't use GTK and uses libui.

Vala has a POSIX profile to write programs without all this glib/gobject dependencies.

PR #342, opened 2025-08-25
darwin: Fix text centering by using constraint width for CTFrame on macOS

kojix2:
### Before

area_breakout-macos

### After

image


kojix2:
I will close this for now. I may send another pull request once the project resumes.

